Issue #274, opened 2014-07-19
Documentation discrepancy regarding etcd access from container


dknell:
According to the docs [Service Discovery with etcd](https://coreos.com/using-coreos/) and [Using docker with etcd](https://coreos.com/using-coreos/docker/), I should be able to run the following from within a container, but I get an error.

```
curl -L http://127.0.0.1:4001/v2/keys/
curl: (7) Failed to connect to 127.0.0.1 port 4001: Connection refused
```

However, if I use the gatway IP (docker0 interface on coreos host) of the container, it works:

```
curl -L http://10.1.42.1:4001/v2/keys/
{"action":"get","node":{"key":"/","dir":true,"nodes":[{"key":"/coreos.com","dir":true,"modifiedIndex":4,"createdIndex":4}]}}
```

I also read [Reading and Writing from Inside a Container](https://coreos.com/docs/distributed-configuration/getting-started-with-etcd/), which states:
To read and write to etcd from within a container you must use the `docker0` interface which you can find in `ip address show`. It's normally `172.17.42.1` and using it is as easy as replacing `127.0.0.1`.

This makes it a little clearer, but the 3 pages I mentioned should do a better job of explaining exactly **_where_** the commands are being run and to which IP addresses should be used. Since there are several layers of virtualization (more so when you are on Vagrant), the details become imperative.

PR #275, opened 2014-07-19
feat(getting-started-etcd): more clarity around reading etcd in a container

robszumski:
Addresses https://github.com/coreos/docs/issues/274


inyoungcho:
The instruction doesn't worked for me.
There is an error in the document, "It's normally `172.17.42.1` and using it is as easy as replacing...."
I had to use the gateway ip of container is 10.240.0.*. to get curl message to work.

PR #412, opened 2015-01-07
Document that build names match git SHAs.

jzelinskie:
I added this line because the only other place it is documented is a question asked via Twitter.

### DIFF
--- a/enterprise-registry/github-build/index.md
+++ b/enterprise-registry/github-build/index.md
@@ -81,4 +81,6 @@ Because tagging is [usually done client side](https://docs.docker.com/userguide/
 
 <img src="{{site.url}}/docs/enterprise-registry/github-build/new-tag.png" class="img-center" alt="Create a new tag"/>
 
+Trying to find a specific image for a commit after it is no longer tagged as latest? The build name (visible in the build history of a repository) will match the commit SHA value.
+
 There is currently no ability to automatically tag GitHub triggered builds.


robszumski:
I dont know if I fully understand the sentence. Can you try rephrasing or explaining it more to me and we can brainstorm?

jzelinskie:
I was basically copying verbatim from twitter. Here's a visual:
screen shot 2015-01-08 at 11 34 02 am

Builds are named after the first 7 digits of the SHA. We also have that value stored as the title of the buildpack.

enterprise-registry/github-build/index.md
```diff
@@ -81,4 +81,6 @@ Because tagging is [usually done client side](https://docs.docker.com/userguide/
 
 <img src="{{site.url}}/docs/enterprise-registry/github-build/new-tag.png" class="img-center" alt="Create a new tag"/>
 
+Trying to find a specific image for a commit after it is no longer tagged as latest? The build name (visible in the build history of a repository) will match the commit SHA value.
```
robszumski: Does this make sense?

"Trying to match GitHub commits with the resulting container build? The build history of your container will show the output of each build, named by the first 7 digits of your GitHub commit SHA value."

jzelinskie:
Apparently this is going to change soon, when we add some more GitHub integrations in the future. I'm not sure if we want to include a warning that it is going to change or just not even mention it at all.

PR #413, opened 2015-01-08
setup: Add new misc settings page for smaller notes.

marineam:
Adds documentation for:
 - /etc/modules-load.d
 - /etc/sysctl.d
 - /etc/motd.d


robszumski:
LGTM.

My only thought is that we get a decent number of questions about tweaking the kernel modules. It might be good to either have that in the title of the doc or have a separate doc altogether. What do you think?

marineam:
If we further expand the document to modprobe.d or what not then yeah, but for now this is fine. We can split it later.

cluster-management/setup/other-settings/index.md
```diff
@@ -0,0 +1,115 @@
...
+to be created first and the file name must end in `.conf`.
+CoreOS 555.0.0 or later is required.
+
+```sh
+mkdir -p /etc/motd.d
+echo "This machine is dedicated to computing Pi" > /etc/motd.d/pi.conf
+```
+
+Or via cloud-config:
+```yaml
```
robszumski: Random FYI: not having a new line before the three backticks breaks our markdown parser.

marineam: Doh, will you fix or should I?

robszumski: I gotcha: https://github.com/coreos/docs/commit/c18ceec21e1acbce2e9ba95b1cd3468e6da59199

PR #415, opened 2015-01-17
Don't call it a "bind mount" when it's not

SpencerBrown:
The mount being done in the example is just a regular device mount, not a bind mount. Remove "bind" from the description.


jonboulle:
lgtm, thanks for the fix!

SpencerBrown:
Johnathan, thanks!

Is the protocol here for me to delete the patch branch on GitHub once it's
merged, or is that taken care of some other way?

Spencer Brown

jonboulle:
@SpencerBrown yes, you'll need to delete it yourself

PR #419, opened 2015-01-22
Explain why docker run --rm isn't used for ExecStart.

### LINKED ISSUE #418: Why do systemd unit examples use docker start/stop/kill instead of docker run --rm?
All of the systemd unit file examples in the documentation use a combination of ExecStartPre, ExecStart, and ExecStop commands to control the Docker container lifecycle. For example, on https://coreos.com/docs/launching-containers/launching/launching-containers-fleet/:

```
[Unit]
Description=MyApp
After=docker.service
Requires=docker.service

[Service]
TimeoutStartSec=0
ExecStartPre=-/usr/bin/docker kill busybox1
ExecStartPre=-/usr/bin/docker rm busybox1
ExecStartPre=/usr/bin/docker pull busybox
ExecStart=/usr/bin/docker run --name busybox1 busybox /bin/sh -c "while true; do echo Hello World; sleep 1; done"
ExecStop=/usr/bin/docker stop busybox1
```

It's never explained why it's necessary (or recommended) to split the container lifecycle into separate commands instead of replacing all of these with `ExecStart=/usr/bin/docker run --rm ...`, which will remove the container automatically when the process ends.

If someone just answers the question here, I'd be happy to open a PR to add it to the docs somewhere. Thanks!

jimmycuadra:
A small addition to explain why separate `ExecStartPre`, `ExecStart`, and `ExecStop` options are preferred to a single `ExecStart` using `docker run --rm`.

Closes #418.

### DIFF
--- a/launching-containers/launching/getting-started-with-systemd/index.md
+++ b/launching-containers/launching/getting-started-with-systemd/index.md
@@ -117,6 +117,8 @@ ExecStopPost=/usr/bin/etcdctl rm /domains/example.com/10.10.10.123:8081
 WantedBy=multi-user.target
 ```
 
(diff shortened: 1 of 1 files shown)


robszumski:
LGTM. Thanks for helping us improve :)

PR #420, opened 2015-01-24
running-coreos: update install script options

robszumski:
Added missing params `-b` and `-v`

cc: @marineam


marineam:
[LGTM](http://www.lgtm.in/i/ajLRzudRp)

PR #430, opened 2015-02-11
Update FLEETCTL_ENDPOINT syntax for unix sockets

AvianFlu:
This syntax was changed in `fleet` 0.9.0 - unix socket paths must now be prepended with `file://`, or else they don't work.


bcwaldon:
@AvianFlu thanks for catching this - do you think it would make a bit more sense to use `unix://` instead? It's opening a unix domain socket, after all.

AvianFlu:
Good point; force-pushed.

bcwaldon:
LGTM /cc @robszumski

robszumski:
Thanks @AvianFlu!

PR #432, opened 2015-02-12
Change ER installation instructions for the new setup tool

josephschorr:


enterprise-registry/initial-setup/index.md
```diff
@@ -8,36 +8,21 @@ weight: 5
...
 
-CoreOS Enterprise Registry requires five components to operate successfully:
+CoreOS Enterprise Registry requires three components to operate successfully:
 
 - A supported database (MySQL, Postgres)
 - A Redis instance (for real-time events)
-- A config.yaml file
-- An SMTP server
 - The Enterprise Registry image
+- (optional) An SMTP server
```
robszumski: I think we need a line here saying that you need to have these 3 running, and our wizard will walk you through setting up the rest. Maybe with a screenshot to hint that you'll need things like the IP/port.

Maybe also mention that you can set up stuff like SMTP later on if desired.

enterprise-registry/initial-setup/index.md
```diff
@@ -48,260 +33,49 @@ sudo docker run -d -p 6379:6379 quay.io/quay/redis
...
+Run the following command, replacing `/local/path/to/the/config/directory` and `/local/path/to/the/storage/directory` with the absolute
+paths to the directories created above:
 
-## Running the Registry
+```
+sudo docker run --restart=always -p 443:443 -p 80:80 --privileged=true -v /local/path/to/the/config/directory:/conf/stack -v /local/path/to/the/storage/directory:/datastorage -d quay.io/coreos/registry
+```
 
-The CoreOS Enterprise Registry is run via a `docker run` call, with the `config` and `storage` being the directories created above.
+Once started, visit: http://yourhost/setup, wait for the page to load (it may take a minute or two) and follow instructions there to setup the enterprise registry.
```
robszumski: A screenshot would be great here

josephschorr:
Added some screenshots. PTAL

robszumski:
Screenshots are nice! LGTM to me.

PR #434, opened 2015-02-15
Fix example command in debugging tools docs

squirly:
In order for a user to use toolbox they need to be in the sudo and docker group.
See [toolbox#Advanced Usage](https://github.com/coreos/toolbox#advanced-usage) for inspiration.


robszumski:
@marineam Does this look good?

marineam:
@squirly I would suggest using sudo/docker both as supplementary groups: `-U -G sudo,docker`

squirly:
@marineam Updated.

robszumski:
@marineam can you take a look at this again?

marineam:
LGTM

PR #436, opened 2015-02-17
reliable container startup

ibukanov:
Using ExecStart to start the container makes systemd to consider the unit started even before the docker client contacted the docker daemon. Thus systemd may start another unit file that depends on this one before the container is created. It makes problematic to run other containers that either link or use the networking namespace or volumes from this container.

The workaround is to start the containers from ExecRunPre using run -d. That command returns only after docker initializes containers allowing reliably to start dependent containers from other units. Then ExecStart just fetches the startup and subsequent logs using docker logs -f.

### DIFF
--- a/launching-containers/launching/getting-started-with-systemd/index.md
+++ b/launching-containers/launching/getting-started-with-systemd/index.md
@@ -44,7 +44,7 @@ The description shows up in the systemd log and a few other places. Write someth
 
 `After=docker.service` and `Requires=docker.service` means this unit will only start after `docker.service` is active. You can define as many of these as you want.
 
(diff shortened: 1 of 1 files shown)


robszumski:
Running containers with `-d` prevents systemd from creating the correct pid hierarchy and monitoring the process. Your `docker logs` usage, while clever, isn't a robust solution.

Your problem really stems from docker's lack of a "pod" or group of containers.

ibukanov:
@robszumski `docker run -d` with `docker logs -f` is no different from `docker run --detach=false`. In both cases systemd monitors the docker client process, not the real container process. However, using `docker run -d` as `ExecStartPre` ensures that the container networking and volumes are ready before systemd considers the unit as ready and starts dependent units allowing reliably order startup and shutdown of a group of linked containers using systemd units. This may not be ideal, but I have found this is good enough to implement the notion of pod reliably.

Also, systemd runs perfectly without

robszumski:
I'm glad this works for you, but I don't think we can document it.

To others: If you find this useful, please comment and we can reconsider and reopen.

PR #441, opened 2015-02-24
Update Vultr documentation.

### LINKED ISSUE #341: Vultr documentation for cloud-config-url
Hi,

it is not clear when and how should be "cloud-config-url" set. Could anyone expand on that? For example, what address should I provide? what it should point to exactly? what would be an example of output? what would change if the option is left invalid?

Thanks

ryneeverett:
1. **Install to disk rather than RAM.** The Vultr page currently explains how to install to RAM, which I believe is unique among the "cloud provider" pages. I can't imagine why one would do this on a cloud server. I can only assume it was done that way because it was the simplest possible installation at the time, but Vultr now provides their own CoreOS image so that is no longer the case.

2. **Use custom `cloud-config`.** Now that Vultr provides a CoreOS image, the only use case for a custom iso is to provide a custom `cloud-config`.

3. **Use Vultr's "Startup Scripts" feature to provide the iPXE.** This feature has been added since the docs and makes the process a lot simpler.

4. **Moved "Cloud Config" above "Choosing a Channel".** You need to know the url to your `cloud-config` before you can write the pxe file.

5. **SSH public key in `cloud-config` rather than pxe.** I actually am not sure if this is necessary, but it works and strikes me as more appropriate.

6. **Remove screenshot.** Screenshot is quite dated. Neither the CoreOS "Operating System" option or the "Startup Scripts" features existed when it was taken. I'd be willing to add an updated screenshot if eye candy is mandatory.

I believe all the other changes are self-explanatory updates.

I think this should close #341.

### DIFF
--- a/running-coreos/cloud-providers/vultr/index.md
+++ b/running-coreos/cloud-providers/vultr/index.md
@@ -19,6 +19,8 @@ The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating
 
 First, you'll need to make a shell script containing your `cloud-config` available at a public URL:
 
+**cloud-config-bootstrap.sh**
+
 ```sh
 #!/bin/bash
 
--- a/running-coreos/cloud-providers/vultr/index.md
+++ b/running-coreos/cloud-providers/vultr/index.md
@@ -34,8 +34,6 @@ Please be sure to check out [Using Cloud-Config]({{site.url}}/docs/cluster-manag
 
 You must add to your ssh public key to your `cloud-config`'s [ssh authorized keys]({{site.url}}/docs/cluster-management/setup/cloudinit-cloud-config/#ssh_authorized_keys) so you'll be able to log in.
 
-Note that the `$private_ipv4` and `$public_ipv4` variables are only supported on Vultr if you have the 'cloud-config-url' option set on your kernel command line. Without this option, you will need to hard code these values into your `cloud-config` file.
-
 ## Choosing a Channel
 
 CoreOS is designed to be [updated automatically]({{site.url}}/using-coreos/updates) with different schedules per channel. You can [disable this feature]({{site.url}}/docs/cluster-management/debugging/prevent-reboot-after-update), although we don't recommend it. Read the [release notes]({{site.url}}/releases) for specific features and bug fixes.
--- a/running-coreos/cloud-providers/vultr/index.md
+++ b/running-coreos/cloud-providers/vultr/index.md
@@ -15,17 +15,20 @@ These instructions will walk you through running a single CoreOS node. This guid
 
 The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating system from Vultr's default offerings. However, most deployments require a custom `cloud-config`, which can only be achieved in Vultr with an iPXE script. The remainder of this article describes this process.
 
-First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script *(be sure to replace the cloud-config URL)*:
+## Cloud-Config
+
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script:
 
 ```sh
 #!/bin/bash
 
-curl -o cloud-config.yaml URL_TO_YOUR_CLOUD_CONFIG
+# Location of your valid cloud-config file.
+URL = "http://example.com/cloud-config.yaml"
+
+curl -o cloud-config.yaml $URL
 sudo coreos-install -d /dev/vda -c cloud-config.yaml
 sudo reboot
 ```
-
-## Cloud-Config
 
 Please be sure to check out [Using Cloud-Config]({{site.url}}/docs/cluster-management/setup/cloudinit-cloud-config).
 
@@ -52,8 +55,11 @@ CoreOS is designed to be [updated automatically]({{site.url}}/using-coreos/updat
 
 <pre>#!ipxe
 
+# Location of your shell script.
+set cloud-config-url http://example.com/cloud-config-bootstrap.sh
+
 set base-url http://alpha.release.core-os.net/amd64-usr/current
-kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=URL_TO_YOUR_SHELL_SCRIPT
+kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=${cloud-config-url}
 initrd ${base-url}/coreos_production_pxe_image.cpio.gz
 boot</pre>
     </div>
@@ -65,8 +71,11 @@ boot</pre>
 
 <pre>#!ipxe
 
+# Location of your shell script.
+set cloud-config-url http://example.com/cloud-config-bootstrap.sh
+
 set base-url http://beta.release.core-os.net/amd64-usr/current
-kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=URL_TO_YOUR_SHELL_SCRIPT
+kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=${cloud-config-url}
 initrd ${base-url}/coreos_production_pxe_image.cpio.gz
 boot</pre>
     </div>
@@ -78,8 +87,11 @@ boot</pre>
 
 <pre>#!ipxe
 
+# Location of your shell script.
+set cloud-config-url http://example.com/cloud-config-bootstrap.sh
+
 set base-url http://stable.release.core-os.net/amd64-usr/current
-kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=URL_TO_YOUR_SHELL_SCRIPT
+kernel ${base-url}/coreos_production_pxe.vmlinuz cloud-config-url=${cloud-config-url}
 initrd ${base-url}/coreos_production_pxe_image.cpio.gz
 boot</pre>
     </div>
--- a/running-coreos/cloud-providers/vultr/index.md
+++ b/running-coreos/cloud-providers/vultr/index.md
@@ -17,17 +17,17 @@ The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating
 
 ## Cloud-Config
 
-First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script:
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and a shell script which you need to create:
 
 ```sh
+URL="http://example.com/cloud-config.yaml"
+
+cat << EOF > ./cloud-config-bootstrap.sh\n
 #!/bin/bash
-
-# Location of your valid cloud-config file.
-URL = "http://example.com/cloud-config.yaml"
-
 curl -o cloud-config.yaml $URL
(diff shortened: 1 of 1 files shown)


running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,27 @@ weight: 10
...
 * You have a public + private key combination generated. Here's a helpful guide if you need to generate these keys: [How to set up SSH keys](https://help.github.com/articles/generating-ssh-keys).
 
-The simplest option to boot up CoreOS is to load a script that contains the series of commands you'd otherwise need to manually type at the command line. This script needs to be publicly accessible (host this file on your own server). Save this script as a text file (.txt extension).
+The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating system from Vultr's default offerings. However, most deployments require a custom `cloud-config`, which can only be achieved in Vultr with an iPXE script. The remainder of this article describes this process.
+
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script *(be sure to replace the cloud-config URL)*:
+
+```sh
+#!/bin/bash
+
```
robszumski: It might make sense to set the location as an env var so that it's super clear what you need to change:

```sh
#!/bin/bash

# Location of your valid cloud-config file
URL = "http://example.com/cloud-config.yaml"

curl -o cloud-config.yaml $URL
sudo coreos-install -d /dev/vda -c cloud-config.yaml
sudo reboot
```

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,27 @@ weight: 10
...
 
 * You have an account at [Vultr.com](https://www.vultr.com).
-* The location of your iPXE script (referenced later in the guide) is located at ```http://example.com/script.txt```
 * You have a public + private key combination generated. Here's a helpful guide if you need to generate these keys: [How to set up SSH keys](https://help.github.com/articles/generating-ssh-keys).
 
-The simplest option to boot up CoreOS is to load a script that contains the series of commands you'd otherwise need to manually type at the command line. This script needs to be publicly accessible (host this file on your own server). Save this script as a text file (.txt extension).
+The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating system from Vultr's default offerings. However, most deployments require a custom `cloud-config`, which can only be achieved in Vultr with an iPXE script. The remainder of this article describes this process.
+
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script *(be sure to replace the cloud-config URL)*:
```
robszumski: Ideally with the change below we can remove this so we don't have too many "notes" or FYI type things floating around across the docs

robszumski: Or rewrite it so it flows more with the paragraph

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,27 @@ weight: 10
...
+
+```sh
+#!/bin/bash
+
+curl -o cloud-config.yaml URL_TO_YOUR_CLOUD_CONFIG
+sudo coreos-install -d /dev/vda -c cloud-config.yaml
+sudo reboot
+```
+
+## Cloud-Config
```
robszumski: It seems like it might make more sense to move this up above line 18.

running-coreos/cloud-providers/vultr/index.md
ryneeverett: Agreed. I'm not too familiar with PXE but it would be nice to use a variable there too if I can figure out the syntax.

running-coreos/cloud-providers/vultr/index.md
ryneeverett: Agreed. Speaking of notes, I was wondering if the following note could be purged:

>Note that the `$private_ipv4` and `$public_ipv4` variables are only supported on Vultr if you have the 'cloud-config-url' option set on your kernel command line. Without this option, you will need to hard code these values into your `cloud-config` file.

But I left it in since I'm not quite sure what it's talking about. (Why would one ever *not* set the 'cloud-config-url' on their kernel command?)

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,30 @@ weight: 10
...
 * You have an account at [Vultr.com](https://www.vultr.com).
-* The location of your iPXE script (referenced later in the guide) is located at ```http://example.com/script.txt```
 * You have a public + private key combination generated. Here's a helpful guide if you need to generate these keys: [How to set up SSH keys](https://help.github.com/articles/generating-ssh-keys).
 
-The simplest option to boot up CoreOS is to load a script that contains the series of commands you'd otherwise need to manually type at the command line. This script needs to be publicly accessible (host this file on your own server). Save this script as a text file (.txt extension).
+The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating system from Vultr's default offerings. However, most deployments require a custom `cloud-config`, which can only be achieved in Vultr with an iPXE script. The remainder of this article describes this process.
+
+## Cloud-Config
+
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and the following shell script:
```
crawford: It might be easier to include the cloud-config in this bootstrap script as a HEREDOC. Just one less step for the user.

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -35,8 +55,11 @@ CoreOS is designed to be [updated automatically]({{site.url}}/using-coreos/updat
 
 <pre>#!ipxe
 
+# Location of your shell script.
+set cloud-config-url http://example.com/cloud-config-bootstrap.sh
```
crawford: Can you mention that "cloud-config-bootstrap.sh" is the above script?

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,30 @@ weight: 10
...
+curl -o cloud-config.yaml $URL
+sudo coreos-install -d /dev/vda -c cloud-config.yaml
+sudo reboot
+```
+
+Please be sure to check out [Using Cloud-Config]({{site.url}}/docs/cluster-management/setup/cloudinit-cloud-config).
+
+You must add to your ssh public key to your `cloud-config`'s [ssh authorized keys]({{site.url}}/docs/cluster-management/setup/cloudinit-cloud-config/#ssh_authorized_keys) so you'll be able to log in.
+
+Note that the `$private_ipv4` and `$public_ipv4` variables are only supported on Vultr if you have the 'cloud-config-url' option set on your kernel command line. Without this option, you will need to hard code these values into your `cloud-config` file.
```
crawford: How are these variables being populated?

ryneeverett: The user needs to replace with the correct value. This was [the suggested approach](https://github.com/coreos/docs/pull/441#discussion_r25226339) of @robszumski.

robszumski: I think @crawford was asking about `$private_ipv4` and `$public_ipv4`

ryneeverett: My bad. I've never dealt with them before, but based on [the CoreOS page on cloud-config](https://coreos.com/docs/cluster-management/setup/cloudinit-cloud-config/#configuration-parameters) it looks like you set them there, making this warning even less coherent.

I'd be in favor of purging the paragraph. I'm not aware of any use case for installing from a custom iso if one isn't defining a `cloud-config` file.

running-coreos/cloud-providers/vultr/index.md
ryneeverett: I'm thinking a HEREDOC will have us covered since that can define the file name.

running-coreos/cloud-providers/vultr/index.md
```diff
@@ -11,10 +11,28 @@ weight: 10
...
+The simplest option to boot up CoreOS is to select the "CoreOS Stable" operating system from Vultr's default offerings. However, most deployments require a custom `cloud-config`, which can only be achieved in Vultr with an iPXE script. The remainder of this article describes this process.
+
+## Cloud-Config
+
+First, you'll need to make two files available at public URL's -- your `cloud-config` file and a shell script which you need to create:
+
+```sh
+URL="http://example.com/cloud-config.yaml"
+
+cat << EOF > ./cloud-config-bootstrap.sh\n
```
crawford: My suggestion was to put the cloud-config.yaml into a HEREDOC instead of fetching it with curl. That way, you only need to host this single script.

ryneeverett: Ah, I see. So something like:

```sh
#!/bin/bash
IFS='' read -r -d '' CLOUD_CONFIG <<'EOF'
# Your Cloud Config:
ssh_authorized_keys:
  - ssh-rsa ...
EOF

sudo coreos-install -d /dev/vda -c $CLOUD_CONFIG
sudo reboot
```

or 

```sh
#!/bin/bash
cat << EOF > cloud-config.yaml
# Your Cloud Config:
ssh_authorized_keys:
  - ssh-rsa ...
EOF

sudo coreos-install -d /dev/vda -c cloud-config.yaml
sudo reboot
```

For some reason whenever I try these I get thrown into a infinite reboot cycle.

ryneeverett: I guess invalid cloud-config's are a known cause of reboot cycles, so there must be something wrong with my HEREDOC.

ryneeverett:
I think something is wrong with this method as the `cloud-config` doesn't seem to get sourced on restart.

ryneeverett:
Oh nevermind, I just got burned by not having `runtime: true` on my `sshd.socket` unit.

ryneeverett:
I actually find that hella confusing because it seems that your `runtime:true` is the *opposite* of passing the `--runtime` flag to `systemctl enable`...but I'm getting pretty off topic here.

running-coreos/cloud-providers/vultr/index.md
ryneeverett: Oh duh, I needed the `#cloud-config` comment.

ryneeverett: I guess that wasn't the only problem. Even when I paste in the exact same `cloud-config` file I use as a standalone, I still get reboot cycles. I even ran the second HEREDOC line locally and diffed the file it creates with my `cloud-config.yaml` and it returned nothing. Stumped.

crawford: Here is the template that I use https://gist.github.com/crawford/f954d17d4c36c747db80

ryneeverett: Thanks. Just switching to your HEREDOC format worked. Maybe it was just a bash-version incompatibility because the previous version worked fine in my local bash.

running-coreos/cloud-providers/vultr/index.md
ryneeverett: done

crawford:
LGTM

robszumski:
@ryneeverett Is this ready to merge?

ryneeverett:
@robszumski Yes, I believe I've addressed all suggestions.

PR #443, opened 2015-02-25
Change logs debugging instructions

josephschorr:


enterprise-registry/log-debugging/index.md
```diff
@@ -11,29 +11,24 @@ weight: 5
...
 
-The aforementioned shell commands are also available in script form at https://github.com/coreos/docs/blob/master/enterprise-registry/log-debugging/tail-gunicorn-logs.sh
+Sign in to a super user account and visit `http://yourregister/superuser` to view the management panel:
+
+<img src="../build-support/superuser.png" class="img-center" alt="Enterprise Registry Management Panel"/>
+
+## View the logs for each service
+
+- Click the logs tab (<span class="fa fa-bug"></span>)
+- To view logs for each service, click the service name at the top. The logs will update automatically.
```
robszumski: Instead of "update", is "steam" a better verb in this context?

robszumski:
Overall I tend to like title case for all of the headings, but I'll leave that up to your personal preference

enterprise-registry/log-debugging/index.md
josephschorr: I prefer "sous vide" myself; better texture than steaming

(and yes, "stream" is probably better)

robszumski:
Are there some common issues that we can list out here?

josephschorr:
Not offhand. The setup tool's verification step seems to handle all the problems we've had reported in the past.

robszumski:
Thats great to hear that we're catching errors before they become a problem. That leaves this doc kind of empty, do you think its worth keeping?

josephschorr:
Yeah, so people still know where to go if they encounter a problem we haven't considered

PR #146, opened 2014-04-09
Added documentation to VirtualBox platform

skarllot:
Added document to run CoreOS on Oracle VirtualBox platform.

Depends on coreos/scripts#220

PR #445, opened 2015-02-27
Virtualbox: Add channel tabs and misc cleanup

robszumski:
Cleans up work done in https://github.com/coreos/docs/pull/146
 - adds the standard channel tabs
 - replaces the hardcoded 298 version number with "stable". I retained a note that specific versions can also be referenced.
 - minor fix for heading capitalization

cc: @marineam


marineam:
[LGTM](http://www.lgtm.in/i/adwPPfKun)

PR #446, opened 2015-02-28
updated with an update-window script that handles reboot in an environme...

xueshanf:
...nt that runs isolated etcd servers and workers.


crawford:
LGTM

PR #447, opened 2015-03-01
Update running-coreos/platforms/vmware/index.md

ThomasWo:
[Documentation update]
The old link (https://developercenter.vmware.com/tool/ovf/3.5.0) pointed to a 404, changing to the tool's generic homepage (https://developercenter.vmware.com/tool/ovf/) which looks exactly the same as the previous link, but doesn't need to be updated every time VMWare pushs an update.


robszumski:
Thanks!

PR #449, opened 2015-03-05
Moved quote in Environment specification

cdwertmann:
Got stung by this when trying to add more than one parameter here. See
https://github.com/coreos/coreos-overlay/issues/1052


crawford:
LGTM

robszumski:
Sorry about that, thanks for the PR.

PR #455, opened 2015-03-05
added locksmith directive to example cloud-config

popsikle:
If you leave locksmith alone, and leave etcd running then your reboot-strategy gets all out of sync due to locksmith only talking to local etcd

### DIFF
--- a/cluster-management/setup/cluster-architectures/index.md
+++ b/cluster-management/setup/cluster-architectures/index.md
@@ -237,10 +237,10 @@ coreos:
   locksmith:
     endpoint: "http://10.0.0.101:4001,http://10.0.0.102:4001,http://10.0.0.103:4001,http://10.0.0.104:4001,http://10.0.0.105:4001"
   units:
+    - name: etcd.service
+      mask: true
     - name: fleet.service
       command: start
-    - name: etcd.service
-      command: stop
   update:
     # CoreUpdate group ID for "Production Workers"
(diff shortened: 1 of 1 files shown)


cluster-management/setup/cluster-architectures/index.md
```diff
@@ -234,9 +234,13 @@ coreos:
   fleet:
     metadata: "role=worker,cabinet=two,disk=spinning"
     etcd_servers: "http://10.0.0.101:4001,http://10.0.0.102:4001,http://10.0.0.103:4001,http://10.0.0.104:4001,http://10.0.0.105:4001"
+  locksmith:
+    endpoint: "http://10.0.0.101:4001,http://10.0.0.102:4001,http://10.0.0.103:4001,http://10.0.0.104:4001,http://10.0.0.105:4001"
   units:
     - name: fleet.service
       command: start
+    - name: etcd.service
```
crawford: etcd doesn't run by default. You shouldn't need this block.

popsikle: i think starting the fleet service starts the etcd service. iirc

crawford: you are correct. I would write the config like this instead:

```yaml
  units:
    - name: etcd.service
      mask: true
    - name: fleet.service
      command: start
```

This prevents etcd from ever running.

popsikle: updated

crawford:
LGTM

PR #458, opened 2015-03-09
fix typo in curl command

jordane:
The unquoted/unescaped question mark breaks this command. Not sure if you guys prefer escaping or quoting, but I chose quoting.


robszumski:
Thanks for catching this!

PR #460, opened 2015-03-10
Mask etcd service in example worker config

databus23:


robszumski:
cc: @marineam or @crawford

crawford:
LGTM

PR #462, opened 2015-03-17
update Using Authentication for a Registry

collinglass:
It was not clear to me when reading the docs that the .dockercfg was generated on docker login and I spent some time trying to write my own. Hopefully this update suffices to inform the next person.

Please leave comments if not merged.

### DIFF
--- a/launching-containers/building/registry-authentication/index.md
+++ b/launching-containers/building/registry-authentication/index.md
@@ -8,7 +8,7 @@ weight: 7
 
 # Using Authentication for a Registry
 
-A json file `.dockercfg` can be created in your home directory that holds authentication information for a public or private docker registry. The auth token is a base64 encoded string: `base64(<username>:<password>)`.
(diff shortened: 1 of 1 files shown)


launching-containers/building/registry-authentication/index.md
```diff
@@ -30,6 +30,9 @@ Here's what an example looks like with credentials for docker's public index and
 }
 ```
 
+Note: The auth token is a base64 encoded string: `base64(<username>:<password>)`.
```
robszumski: I'd like to avoid having to say "note", over time it will make it harder to read as you have more notes. What if you said:

> If you want to populate these values without running `docker login`, the auth token is a base64 encoded string: `base64(<username>:<password>)`.

collinglass:
Thanks for the feedback, I made the suggested update.

robszumski:
Thanks!

collinglass:
I noticed that the base64:(<username>:<password>) is not displaying <username> and <password>. It should be in code snippet tags ```base64:(<username>:<password>)```.

PR #463, opened 2015-03-17
flannel: instructions for Vagrant

alban:
In addition to the change requested in https://github.com/coreos/coreos-vagrant/pull/214, the documentation could explain how to make sure flannel uses the correct network interface on Vagrant.

/cc @eyakubovich


eyakubovich:
LGTM, thanks!

PR #465, opened 2015-03-21
Fix markdown

gdoteof:
Missing line break between embedded snipped and previous paragraph.  This is what this renders like outside of github:

![broken formatting] (http://i.imgur.com/NNhKTvz.png)


robszumski:
Thanks!

PR #466, opened 2015-03-26
flannel: mention flannel_docker_opts.env; more info on deps requirements

eyakubovich:


robszumski:
LGTM.

Maybe consider an example unit with those options in it so that you can copy and paste it to modify?

PR #471, opened 2015-04-02
Hint metadata and global units can be combined

bitglue:
This never occurred to me after months of wondering about how I might schedule
units to take advantage of AWS auto-scaling. It's obvious in retrospect, but
there's a mental speed-bump: how can a unit be "global" if it runs only on some
subset, according to metadata? This hint should be sufficient to avoid that
mental block.

### DIFF
--- a/launching-containers/launching/launching-containers-fleet/index.md
+++ b/launching-containers/launching/launching-containers-fleet/index.md
@@ -294,6 +294,8 @@ MachineMetadata=platform=cloud
 MachineMetadata=region=east
 ```
 
+It's also possible to restrict a global unit by metadata, thus scheduling a unit on all machines matching some requirements.
+
 #### More Information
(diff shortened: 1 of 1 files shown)


robszumski:
Great point. Can you add an example like the ones above this sentence. Maybe like this:

```
[X-Fleet]
Global=true
MachineMetadata=region=east
```

ElijahCaine:
@bitglue I'd love to get this merged soon. Do you think you'd be able to fix it up and add an example (hopefully tested) to make that happen?
If you're busy I can take ownership of this, or we can close it if the change is outdated.

bitglue:
Unfortunately I don't have a CoreOS cluster anymore so I can't easily test :(

zbwright:
Thanks for your work. Fleet is now deprecated. Closing this PR.

PR #473, opened 2015-04-05
Make networkd unit more permanent

jedsmith:
If you work from this example with `runtime: true`, the unit is created
under `/run`. Normally, this would be fine for most uses, but on Amazon
DHCP will configure both interfaces. To prevent this, the unit files
must be written permanently because DHCP runs before cloud-config.

Dropping `runtime: true` here will have no effect on most installs, but
will allow static routing and preventing asymmetric routes on Amazon.

I only changed one line. The ending was touched by GitHub's editor.

### DIFF
--- a/cluster-management/setup/network-config-with-networkd/index.md
+++ b/cluster-management/setup/network-config-with-networkd/index.md
@@ -54,7 +54,6 @@ coreos:
     - name: fleet.service
       command: start
     - name: 00-eth0.network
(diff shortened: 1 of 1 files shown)


crawford:
When runtime is set to false, it makes it such that the first boot is a special case. As a result, you cannot just boot a machine with a cloud-config and go. You'll need to reboot at least once before the network will work properly.

PR #474, opened 2015-04-07
Correct env file name

kelcecil:
I believe Item 5 in the sequence of events in the flannel-config should be /run/flannel_docker_opts.env according to the flanneld.service file at https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel/files/flanneld.service and to match up with the filename in Item 7 in the same sequence.


robszumski:
@eyakubovich Can you review this?

eyakubovich:
LGTM

PR #478, opened 2015-04-11
Correct 'Ubuntu' to 'minimal busybox'

eadmund:
The busybox container in the examples is not an actual Ubuntu installation; this change updates the text to reflect this.


robszumski:
Thanks!

PR #479, opened 2015-04-14
cluster-management: remove obsolete fleetctl.sh script

bcwaldon:
This is no longer necessary with fleet v0.10.0. We should probably wait to merge this PR until v0.10.0 is available on Stable, though, as fleetctl maintains backwards compatibility with this behavior. It's basically going to be a NOP moving forward.


robszumski:
Handled in another PR

PR #482, opened 2015-04-15
Made comment about Description more clear.

amgrice:
Was going through documentation and was confused for a second. Made the Description a noticeable attribute, not a passing comment.


robszumski:
Thanks!

PR #483, opened 2015-04-16
ntp: report 'ntpd' as disabled and use coreos ntp.org pool

marineam:
The ntpd service is enabled in /usr instead of /etc and therefore
systemd always reports it as "disabled". Also we should not be telling
users to use the generic pool, ntp.org expects us to use our own coreos
pool.


crawford:
LGTM

PR #484, opened 2015-04-19
ec2: Clarify that two exemptions are needed

### LINKED ISSUE #453: Error in Security Group Instructions
Step 5 says to create a rule for the "7001, 4001" port range which is incorrect.

It should say to create a rule for the 4001 port and a second rule for the 7001 port.

jedsmith:
Be specific here that two rules need to be created for users unfamiliar
with how "port range" works in security groups.

Fixes #453


robszumski:
LGTM

PR #488, opened 2015-04-25
Adding Exoscale deployment instruction for CoreOS

retrack:
This PR details how to deploy a CoreOS instance on Exoscale using:

* API
* Web portal

and also explains Security groups and Cloud-config basic information.


robszumski:
It looks like a ton of older commits from 2014 got sucked into this PR. Can you get rid of those?

I added a few comments inline as well.

retrack:
Sorry for the dirty work. I will close the PR and submit a new one with the necessary corrections.

PR #492, opened 2015-04-29
Update instructions for VMware install to disk

Daveforprojects:
In the current page ( https://coreos.com/docs/running-coreos/bare-metal/installing-to-disk/ ) it is not clear how the ISO can be used to install to disk. In particular, users need to be aware that they need to reference a cloud-config.yml file to install login credentials. 

This patch updates the docs to clarify that:
-- the iso will install to disk
-- users need to sudo to root to execute the coreos-install script
-- users need to include a reference to cloud-config.yml


running-coreos/bare-metal/installing-to-disk/index.md
```diff
@@ -15,14 +15,27 @@ weight: 7
...
 coreos-install -d /dev/sda
 ```
 
+If you are using the ISO with VMware, first sudo to root: 
+
+```sh
+sudo su - root
+```
+
+Then install as you would with the PXE booted system, but be sure to include user information, especially an ssh key, in a Cloud-Config file, or else you will not be able to log into your CoreOS instance.
```
robszumski: capitalize SSH

robszumski: Can you link to the cloud-config header below: `[cloud-config](#cloud-config)`

Daveforprojects: Rob -

Thanks for your comments. I've made the changes and pushed them to my repo.
I'm new to the process, shall I make an updated pull request or is that
sufficient?

Dave

robszumski:
Thanks!

PR #495, opened 2015-05-01
running-coreos: align DO examples on stable channel

bcwaldon:


crawford:
LGTM

robszumski:
Whoops, I just changed part of this without looking closely. Thanks for spotting this.

PR #496, opened 2015-05-01
launching-containers: explain fleet machine failure

robszumski:
Related to https://github.com/coreos/fleet/issues/1156


robszumski:
@olalonde Does this cover the information you were looking for on coreos/fleet#1156?

PR #497, opened 2015-05-04
modifying cloud-config after launch on Exoscale

retrack:
A detailed section on the cloud-config chapter about modifying the contents of cloud-config for an instance that was already created.


robszumski:
Thanks, LGTM.

PR #498, opened 2015-05-12
Fix race condition with device availability

rdark:
Occasionally for some reason ephemeral block devices seem to take a little while to become available.

This fixes it.


marineam:
Should include a requires as well.

rdark:
@marineam - added + squashed into previous commit

marineam:
Sorry for the delay. Thanks!

PR #499, opened 2015-05-19
ntp: update to cover systemd-timesyncd

marineam:
timesyncd is now the default so we need to document configuring it as
well as ntpd. Some people will need to switch back to ntpd as well.
Also note that UTC is the default.

Fixes https://github.com/coreos/bugs/issues/363

### DIFF
--- a/cluster-management/setup/configuring-date-and-timezone/index.md
+++ b/cluster-management/setup/configuring-date-and-timezone/index.md
@@ -8,26 +8,95 @@ weight: 7
 
 # Configuring Date and Timezone
 
-NTP is used to to keep clocks in sync across machines in a CoreOS cluster. The ntpd service is responsible for keeping each machines local clock in sync with a configured set of time servers. The services will automatically start by default. To check if the ntpd service is running, run the follow command:
+NTP is used to to keep clocks in sync across machines in a CoreOS cluster.
+CoreOS [681.0.0][681.0.0] uses [systemd-timesyncd][systemd-timesyncd] as the
+default NTP client, prior to that [ntpd][ntp.org] was used. Depending on the
+your version of CoreOS one of the two services will automatically start. To
+check which service is running, run the follow command:
 
 ```
-systemctl status ntpd
-ntpd.service - Network Time Service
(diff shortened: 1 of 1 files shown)


crawford:
LGTM

cluster-management/setup/configuring-date-and-timezone/index.md
```diff
@@ -8,7 +8,11 @@ weight: 7
...
 
-NTP is used to to keep clocks in sync across machines in a CoreOS cluster. The ntpd service is responsible for keeping each machines local clock in sync with a configured set of time servers. The services will automatically start by default. To check if the ntpd service is running, run the follow command:
+NTP is used to to keep clocks in sync across machines in a CoreOS cluster.
+CoreOS [681.0.0][681.0.0] uses [systemd-timesyncd][systemd-timesyncd] as the
+default NTP client, prior to that [ntpd][ntp.org] was used. Depending on the
+your version of CoreOS one of the two services will automatically start. To
+check if the ntpd service is running, run the follow command:
 
 ```
 systemctl status ntpd
```
robszumski: If your machine is using `systemd-timesyncd`, do you still ask for the `ntpd` unit?

cluster-management/setup/configuring-date-and-timezone/index.md
```diff
@@ -20,16 +24,72 @@ ntpd.service - Network Time Service
...
+```
+
+```yaml
+#cloud-config
+
+coreos:
+  units:
+    - name: systemd-timesyncd.service
+      command: stop
+      mask: true
```
robszumski: Can we add a comment on to here that describes what its doing? I'm thinking of a shorter version of this:

```
mask: true    #mask will ensure that this unit isn't started if it was previously enabled
```

marineam: Added a note below instead.

cluster-management/setup/configuring-date-and-timezone/index.md
marineam: Updated to include both units.

robszumski:
LGTM

PR #508, opened 2015-06-18
cluster-arch: switch to etcd2 service

xiang90:
The next step is to introduce proxy into the central service machines setup and simplify fleet initial configuration.

### DIFF
--- a/cluster-management/setup/cluster-architectures/index.md
+++ b/cluster-management/setup/cluster-architectures/index.md
@@ -111,10 +111,13 @@ Here's the cloud-config for the etcd machine:
 #cloud-config
 
 coreos:
-  etcd:
-    addr: 10.0.0.101:4001
+  etcd2:
+    advertise-client-urls: http://10.0.0.101:2379
+    # listen on both the official ports and the legacy ports
+    # legacy ports can be omitted if your application doesn't depend on them
+    listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
   units:
-    - name: etcd.service
+    - name: etcd2.service
       command: start
     - name: 00-eth0.network
       runtime: true
@@ -140,7 +143,7 @@ Look how simple the cloud-config becomes:
 coreos:
   fleet:
     # replace this IP
-    etcd_servers: "http://10.0.0.101:4001"
+    etcd_servers: "http://10.0.0.101:2379"
   units:
     - name: etcd.service
       mask: true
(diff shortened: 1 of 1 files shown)


cluster-management/setup/cluster-architectures/index.md
```diff
@@ -140,7 +143,7 @@ Look how simple the cloud-config becomes:
 coreos:
   fleet:
     # replace this IP
-    etcd_servers: "http://10.0.0.101:4001"
+    etcd_servers: "http://10.0.0.101:2379"
```
robszumski: Don't we want to get rid of this and use the proxy?

cluster-management/setup/cluster-architectures/index.md
```diff
@@ -138,12 +144,13 @@ Look how simple the cloud-config becomes:
 #cloud-config
 
 coreos:
+  etcd2:
+    proxy: on 
+    listen-client-urls: http://localhost:2379
```
robszumski: `-advertise-client-urls is required when -listen-client-urls is set explicitly`

xiang90: when you are running proxy mode, you do not need to do this. there was a bug in 2.0.11 and we fixed it in 2.0.12. coreos skipped etcd 2.0.11

cluster-management/setup/cluster-architectures/index.md
```diff
@@ -138,12 +144,13 @@ Look how simple the cloud-config becomes:
...
+    initial-cluster: etcdserver=http://10.0.0.101:2380
   fleet:
-    # replace this IP
-    etcd_servers: "http://10.0.0.101:4001"
+    etcd_servers: "http://localhost:2379"
   units:
-    - name: etcd.service
-      mask: true
```
robszumski: keep this add start it

cluster-management/setup/cluster-architectures/index.md
```diff
@@ -138,12 +144,15 @@ Look how simple the cloud-config becomes:
...
+    - name: etcd2.service
```
xiang90: @robszumski I should start etcd2 instead, right? or I still need to mask etcd?

robszumski: yes, it will need to be etcd2. 

@crawford is it safer to mask the other etcd unit here? or should we remove?

cluster-management/setup/cluster-architectures/index.md
robszumski: I was running into that on beta. 

@crawford Are we going to ship 2.0.12 to all channels soon? Customers are going to run into this error otherwise.

robszumski: Er, nevermind, it was alpha.

robszumski:
Tested this and it looks good to go. After merging I'll submit another PR to clean up some outdated info in the images.

PR #517, opened 2015-06-23
docs: only expose the new etcd server port

crawford:
It really only makes sense to expose the legacy client port since new
clusters will all use the new server port to communicate.


robszumski:
LGTM

PR #523, opened 2015-07-02
Removed long ssh-rsa keys, updated iPXE and PXE docs, fixed links to …

kayrus:
…"config_drive" docs

/cc @brianredbeard
/cc @robszumski

### DIFF
--- a/cluster-management/setup/adding-users/index.md
+++ b/cluster-management/setup/adding-users/index.md
@@ -25,7 +25,7 @@ users:
       - sudo
       - docker
     ssh-authorized-keys:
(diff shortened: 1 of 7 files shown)


robszumski:
LGTM but I'd love to get more eyes on the iPXE stuff.

marineam:
LGTM

marineam:
actually, if we are dropping the rootfstype=btrfs in the doc we can drop the tmpfs one too since it doesn't make much sense without the alternative option.

kayrus:
@robszumski specially for you I've added simple nginx config to host cloud-config files =)

marineam:
@robszumski  I'm going to go ahead and merge, @ionave has related changes and want to avoid conflicts :)

PR #524, opened 2015-07-03
etcd doc was updated with exec-watch command

kayrus:
I've also moved "--recursive" option before the key. As it doesn't work with exec-watch:

```
etcdctl exec-watch /foo-service --recursive -- echo foo
```

It should be:

```
etcdctl exec-watch --recursive /foo-service -- echo foo
```


kayrus:
@brianredbeard @robszumski

distributed-configuration/getting-started-with-etcd/index.md
```diff
@@ -145,6 +145,34 @@ $ curl -L http://127.0.0.1:4001/v2/keys/foo-service?wait=true\&recursive=true
...
+
+```
+
+In the other window, let's imagine a new container has started and announced itself to the proxy by running:
+
+```sh
+$ etcdctl set /foo-service/container3 localhost:2222
+localhost:2222
+```
+
```
robszumski: I was confused by having both the etcdctl and curl examples. I think we should remove the curl one since we cover that earlier in the doc.

robszumski:
LGTM other than the one comment

kayrus:
Done

robszumski:
LGTM

robszumski:
I made this unmergable with https://github.com/coreos/docs/pull/529. If you have a sec to update it that would be great...if not it's on my list of stuff to do, but that list is long :)

All the same docs exist, the hierarchy was just flattened.

kayrus:
done

robszumski:
@kayrus Thanks, you're the best!

PR #526, opened 2015-07-07
Updated etcd2 port in examples and added NFS mount example

kayrus:
@robszumski @brianredbeard

### DIFF
--- a/cluster-management/setup/mounting-storage/index.md
+++ b/cluster-management/setup/mounting-storage/index.md
@@ -108,6 +108,36 @@ coreos:
 
 Note the declaration of `ConditionPathExists=!/var/lib/docker.btrfs`. Without this line, systemd would reformat the btrfs filesystem every time the machine starts.
 
+## Mounting NFS shares
+
+This example will show how you can mount NFS share. First of all we will enable rpc-statd for NFS host monitoring. Then we will mount `/var/www` share into local `/var/www` path. Please pay attention on the unit name `var-www.mount` it should correspend to mount path `/var/www`.
+
+```yaml
+#cloud-config
+coreos:
+  units:
+    - name: rpc-statd.service
+      command: start
+      enable: true
+    - name: var-www.mount
+      command: start
+      content: |
+        [Mount]
+        What=nfs.example.com:/var/www
+        Where=/var/www
+        Type=nfs
+```
+
+If you would like to add NFS mount as dependency for certain unit you have to use these options:
+
+```yaml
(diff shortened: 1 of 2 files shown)


quickstart/index.md
```diff
@@ -74,13 +74,13 @@ The API is easy to use. From a CoreOS machine, you can simply use curl to set an
 Set a key `message` with value `Hello world`:
 
 ```sh
-curl -L http://127.0.0.1:4001/v2/keys/message -XPUT -d value="Hello world"
+curl -L http://127.0.0.1:2379/v2/keys/message -XPUT -d value="Hello world"
```
robszumski: Looks like we completely forgot to update the quick start to use etcd2. Can you also change the cloud-config exmaple above to use etcd2? It's currently:

```
coreos:
  units:
    - name: etcd.service
```

robszumski: Hmm, I wonder why these are curl to begin with, instead of etcdctl. I wonder if we should make one of them etcdctl and keep one curl to show that you don't have to use your tool.

In order to do that we'd probably need to add a sentence about what etcdctl is and then that you can also use any regular http stuff like curl.

"etcdctl is a command line interface to etcd that is preinstalled on coroes. to set a key, blah blah:"

cluster-management/setup/mounting-storage/index.md
```diff
@@ -108,6 +108,26 @@ coreos:
...
+      enable: true
+    - name: var-www.mount
+      command: start
+      content: |
+        [Mount]
+        What=nfs.example.com:/var/www
+        Where=/var/www
+        Type=nfs
+```
+
```
robszumski: Since you had issues with stuff depending on your example mount, should we add a note about that here?

"If you have units that dep. on this, you might need to X and Y"

kayrus:
Done

cluster-management/setup/mounting-storage/index.md
```diff
@@ -108,6 +108,36 @@ coreos:
...
+If you would like to add NFS mount as dependency for certain unit you have to use these options:
+
+```yaml
+[Unit]
+After=var-www.mount
+Requires=var-www.mount
+```
+
+If mount point could not be mounted unit will not start.
```
robszumski: If **the** mount point **can't be** mounted **correctly**, **the dependent** unit will not start.

kayrus:
Done

robszumski:
LGTM

PR #529, opened 2015-07-15
*: shuffle and reorganize

robszumski:
Move documents into subfolders by project, instead of deeply nested categories.

Remove the front matter to be stored elsewhere since contributors don't need to mess with that.


crawford:
LGTM

robszumski:
Rebased since I missed a few image paths

crawford:
LGTM

PR #534, opened 2015-07-17
Added systemd timers doc

kayrus:
/cc @robszumski

### DIFF
--- a/os/scheduling-tasks-with-systemd-timers.md
+++ b/os/scheduling-tasks-with-systemd-timers.md
@@ -1,0 +1,71 @@
+# Scheduling Tasks with systemd Timers
+
+CoreOS uses systemd timers (`cron` replacement) to schedule tasks. Here we will show you how you can schedule a periodic job.
+
+Let's create an alternative for this `crontab` job:
+
+```
+*/10 * * * * /usr/bin/date >> /tmp/date
+```
+
+Timers work directly with services' units. So we have to create `/etc/systemd/system/date.service` first:
+
+```
+[Unit]
+Description=Prints date into /tmp/date file
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sh -c '/usr/bin/date >> /tmp/date'
+```
+
+Then we have to create timer unit with the same name but with `*.timer` suffix `/etc/systemd/system/date.timer`:
+
+```
+[Unit]
+Description=Run date.service every 10 minutes
+
+[Timer]
+OnCalendar=*:0/10
+```
+
+
+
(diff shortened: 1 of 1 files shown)


os/task-scheduler-with-systemd-timers.md
```diff
@@ -0,0 +1,63 @@
+# Task scheduler with systemd timers
```
robszumski: title case

what about: "Scheduling Tasks with systemd Timers"?

os/task-scheduler-with-systemd-timers.md
```diff
@@ -0,0 +1,63 @@
...
+
+CoreOS doesn't use `cron` scheduling daemon. The replacement is `systemd` timer unit. Here we will show you how you can schedule a periodic job.
```
robszumski: I feel like this leads off with a negative. Should we say "CoreOS uses systemd timers to..." and then say that its similar to cron, but with better syntax.

os/task-scheduler-with-systemd-timers.md
```diff
@@ -0,0 +1,63 @@
...
+```
+[Unit]
+Description=Prints date into /tmp/date file
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sh -c '/usr/bin/date >> /tmp/date'
+```
+
+Then we have to create timer unit `/etc/systemd/system/date.timer`:
```
robszumski: These have to be named the same, correct? If so, we need to mention that.

os/task-scheduler-with-systemd-timers.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+[Unit]
+Description=Run date.service every 10 minutes
+
+[Timer]
+OnCalendar=*:0/10
+```
+
+This config will run `date.service` every 10 minutes. You can also list all timers enabled in your system using `systemctl list-timers` command or `systemctl list-timers --all` to list all timers. Run `systemctl start date.timer` to enable timer.
```
robszumski: Nice job mentioning all this stuff!

os/task-scheduler-with-systemd-timers.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Cloud Config
```
robszumski: Cloud-Config

kayrus:
Done!

robszumski:
Nice looks great. This is cool, I didn't realize this:

> You can also create timer with different name, i.e. `task.timer`. In this case you have specify service unit name:

PR #548, opened 2015-07-28
Fixed links, formatting

kayrus:
/cc @robszumski


robszumski:
LGTM

Side note: there are probably a lot of out of date links, but we have redirects in place for them. We can slowly clean them up over time...

PR #549, opened 2015-07-29
os: update vmware guide

crawford:
Users should be using the standard VMware images. This suggests that
they use the OVF image so it can be directly imported into ESXi.

### DIFF
--- a/os/booting-on-vmware.md
+++ b/os/booting-on-vmware.md
@@ -1,71 +1,68 @@
 # Running CoreOS on VMware
 
-CoreOS is currently in heavy development and actively being tested.
-These instructions will walk you through running CoreOS on VMware Fusion or ESXi.
-If you are familiar with another VMware product you can use these instructions as a starting point.
+These instructions will walk you through running CoreOS on VMware Fusion or
+ESXi. If you are familiar with another VMware product you can use these
+instructions as a starting point.
 
 ## Running the VM
 
-These steps will download the VMware image and extract the zip file. After that
-you will need to launch the `coreos_developer_vmware_insecure.vmx` file to create a VM.
-
 ### Choosing a Channel
 
-CoreOS is released into stable, alpha and beta channels. Releases to each channel serve as a release-candidate for the next channel. For example, a bug-free alpha release is promoted bit-for-bit to the beta channel.
+CoreOS is released into alpha, beta, and stable channels. Releases to each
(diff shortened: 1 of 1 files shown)


crawford:
I haven't yet tested this guide.

os/booting-on-vmware.md
```diff
@@ -1,71 +1,68 @@
...
+The channel is selected based on the URLs below. Simply replace `alpha` with
+`beta` or `stable`. Read the [release notes][release notes] for specific
+features and bug fixes in each channel.
 
 ```sh
-curl -LO http://alpha.release.core-os.net/amd64-usr/current/coreos_production_vmware_insecure.zip
-unzip coreos_production_vmware_insecure.zip -d coreos_production_vmware_insecure
-cd coreos_production_vmware_insecure
-open coreos_production_vmware_insecure.vmx
+curl -LO http://alpha.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ovf
```
robszumski: Do you want this to change based on the tab selected? If so, we need to add the tabs. If not, we should probably default to stable.

robszumski:
Text looks good to me other than the tab question.

os/booting-on-vmware.md
crawford: I'll default to stable.

os/booting-on-vmware.md
```diff
@@ -1,71 +1,68 @@
...
+The channel is selected based on the URLs below. Simply replace `stable` with
+`alpha` or `beta`. Read the [release notes][release notes] for specific
+features and bug fixes in each channel.
 
 ```sh
-curl -LO http://alpha.release.core-os.net/amd64-usr/current/coreos_production_vmware_insecure.zip
-unzip coreos_production_vmware_insecure.zip -d coreos_production_vmware_insecure
-cd coreos_production_vmware_insecure
-open coreos_production_vmware_insecure.vmx
+curl -LO http://stable.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ovf
```
marineam: .ova not .ovf

crawford: oops, both are needed

marineam: no, .ova includes the ovf and vmdk. If you fetch the ovf you do need to fetch the vmdk.

PR #552, opened 2015-07-29
os: remove executable bit from booting-on-vultr.md

mischief:


robszumski:
LGTM, this must have been set by the contributor.

PR #555, opened 2015-08-04
Added udev example

kayrus:
/cc @robszumski 
Rob, can you suggest another udev example?

### DIFF
--- a/os/using-systemd-and-udev-rules.md
+++ b/os/using-systemd-and-udev-rules.md
@@ -1,6 +1,6 @@
 # Using systemd and udev Rules 
 
-In our example we will use libvirt VM. First of all we have to create systemd unit file `/etc/systemd/system/test.service`:
+In our example we will use libvirt VM with CoreOS and run systemd unit on disk attach event. First of all we have to create systemd unit file `/etc/systemd/system/device-attach.service`:
 
 ```
 [Service]
@@ -8,9 +8,11 @@ Type=oneshot
 ExecStart=/usr/bin/echo 'device has been attached'
 ```
 
+This unit file will be triggered by our udev rule.
+
 Then we have to start `udevadm monitor --environment` to monitor kernel events.
 
-Once you've attached virtio libvirt device (i.e. `virsh attach-disk coreos /dev/VG/test vdc`) you'll see similar udev output:
+Once you've attached virtio libvirt device (i.e. `virsh attach-disk coreos /dev/VG/test vdc`) you'll see similar `udevadm` output:
 
 ```
 UDEV  [545.954641] add      /devices/pci0000:00/0000:00:18.0/virtio4/block/vdb (block)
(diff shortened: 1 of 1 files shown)


robszumski:
I don't know a ton about udev rules so I found this entire doc hard to follow. Adding some comments inline.

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
+# Using systemd and udev Rules 
+
+In our example we will use libvirt VM. First of all we have to create systemd unit file `/etc/systemd/system/test.service`:
```
robszumski: Libvirt is one of our less popular platforms. Is there a way we can make this guide generic or use a different platform as the example?

robszumski: Why are we creating this unit? What does it do?

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/echo 'device has been attached'
+```
+
+Then we have to start `udevadm monitor --environment` to monitor kernel events.
+
+Once you've attached virtio libvirt device (i.e. `virsh attach-disk coreos /dev/VG/test vdc`) you'll see similar udev output:
```
robszumski: Where do I see this output?

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
...
+MAJOR=254
+MINOR=16
+SEQNUM=1327
+SUBSYSTEM=block
+SYSTEMD_WANTS=test.service
+TAGS=:systemd:
+USEC_INITIALIZED=545954447
+```
+
+Accoring to text beyond udev rule should look this way:
```
robszumski: "according"

robszumski: According to what text?

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+ACTION=="add", SUBSYSTEM=="block", TAG+="systemd", ENV{SYSTEMD_WANTS}="test.service"
+```
+
+Now when we use this command `virsh attach-disk coreos /dev/VG/test vdc` on host machine, we should see `device has been attached` message inside CoreOS node.
```
robszumski: So the test.service unit triggers the disk to be attached?

os/using-systemd-and-udev-rules.md
robszumski: Can we rename this unit to something like `disk-notify` or `device-attach` or `device-notify`?

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Cloud-Config example
+
+Cloud-Config example should look this way:
```
robszumski: "To use the unit and udev rule with cloud-config, modify this example as needed:"

os/using-systemd-and-udev-rules.md
```diff
@@ -0,0 +1,74 @@
...
+        Description=Notify about attached device
+
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/bin/echo 'device has been attached'
+```
+
+## Another Examples
+
+For another real examples, check out these documents:
```
robszumski: How do these relate?

kayrus:
Updated

robszumski:
Let's just get this in. LGTM and we can tweak if needed.

PR #556, opened 2015-08-07
kubernetes: add pod, service and RC overview

robszumski:
Can you take a look at this @kelseyhightower?


robszumski:
FYI: for some reason Github won't show the SVGs inline :/

PR #561, opened 2015-08-13
os: Fixing quoting on example for Packet

stresler:


stresler:
more incoming in a single PR.

PR #565, opened 2015-08-19
ER: fix broken URL to download logs script

jzelinskie:


robszumski:
Whoops, thanks. LGTM.

There are only a few circumstances that our redirects don't cover, and this is one of them.

PR #567, opened 2015-08-20
Fix issue when network is not yet stopped but cloudinit tries to star…

kayrus:
…t systemd-networkd. "stop" command is not necessary here.

/cc @robszumski

### DIFF
--- a/os/network-config-with-networkd.md
+++ b/os/network-config-with-networkd.md
@@ -90,7 +90,7 @@ coreos:
         ExecStart=/usr/bin/ip link set eth0 down
         ExecStart=/usr/bin/ip addr flush dev eth0
     - name: systemd-networkd.service
-      command: start
+      command: restart
 ```
 
 ## Turn Off DHCP on specific interface


os/network-config-with-networkd.md
```diff
@@ -70,8 +70,6 @@ By default, even if you've already set a static IP address and you have a workin
 
 coreos:
   units:
-    - name: systemd-networkd.service
-      command: stop
```
robszumski: Are you sure? The network can sometimes be up before cloudinit runs. I'm assuming this is here for a reason.

cc @crawford

kayrus: Yes, I've checked that several times. Sometimes node's network is up, sometimes - not.

crawford: networkd needs to be stopped before we take down the interfaces. Otherwise, networkd will bring them right back up. This whole this is a shitty pile of hacks and we should really focus on Ignition instead.

kayrus: Ok, if we keep "stop" - we should use "restart" at the end anyway.

crawford: That's fine, but why? The race condition?

kayrus: Yep

kayrus:
Any news?

crawford:
Sorry, I thought I already commented on this.

LGTM

crawford:
Please clean up the commit message before merging.

kayrus:
Done

PR #571, opened 2015-08-28
Fixed cloud-config examples for platforms

kayrus:
* Added complete example for bare-metal setup
* Added additional info for Vagrant
* Tiny fixes

/cc @robszumski

### DIFF
--- a/os/booting-on-vagrant.md
+++ b/os/booting-on-vagrant.md
@@ -50,20 +50,31 @@ coreos:
     listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
     listen-peer-urls: http://$private_ipv4:2380
   fleet:
-      public-ip: $public_ipv4
+    public-ip: $public_ipv4
+  flannel:
+    interface: $public_ipv4
   units:
     - name: etcd2.service
       command: start
     - name: fleet.service
       command: start
+    - name: flanneld.service
+      drop-ins:
+      - name: 50-network-config.conf
+        content: |
+          [Service]
+          ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{ "Network": "10.1.0.0/16" }'
+      # Uncomment line above if you want to use flannel in your installation.
+      # command: start
 ```
 
 The `$private_ipv4` and `$public_ipv4` substitution variables are fully supported in cloud-config on Vagrant. They will map to the first statically defined private and public networks defined in the Vagrantfile.
 
+There is no need to add an SSH key since Vagrant will automatically generate and use it's own SSH key. Any keys added will be overwritten.
+
 Your Vagrantfile should copy your cloud-config file to `/var/lib/coreos-vagrant/vagrantfile-user-data`. The provided Vagrantfile is already configured to do this. `cloudinit` reads `vagrantfile-user-data` on every boot and uses it to create the machine's user-data file.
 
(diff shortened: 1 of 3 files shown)


os/booting-on-vagrant.md
```diff
@@ -50,20 +50,56 @@ coreos:
...
+      - name: 50-network-config.conf
+        content: |
+          [Service]
+          ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{ "Network": "10.1.0.0/16" }'
+      command: start
 ```
 
 The `$private_ipv4` and `$public_ipv4` substitution variables are fully supported in cloud-config on Vagrant. They will map to the first statically defined private and public networks defined in the Vagrantfile.
 
+Also please note that you should not put your public SSH keys in cloud-config as Vagrant will overwrite them with its own automatically generated key.
```
robszumski: Let's rewrite this to not start with "also". What about:

"There is no need to add an SSH key since Vagrant will automatically generate and use it's own SSH key. Any keys added will be overwritten."

os/booting-on-vagrant.md
```diff
@@ -50,20 +50,56 @@ coreos:
...
-      public-ip: $public_ipv4
+    public-ip: $public_ipv4
+  flannel:
+    interface: $public_ipv4
   units:
     - name: etcd2.service
       command: start
     - name: fleet.service
       command: start
+    - name: flanneld.service
```
robszumski: Not all users need to use or configure flannel. Should we document this as a default?

kayrus:
Updated

os/booting-on-vagrant.md
```diff
@@ -50,20 +50,57 @@ coreos:
...
+      drop-ins:
+      - name: 50-network-config.conf
+        content: |
+          [Service]
+          ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{ "Network": "10.1.0.0/16" }'
+      # Substitute "stop" to "start" if you want to use flannel in your installation.
+      command: stop
```
robszumski: Hmm, I wonder if this is the correct solution. Thoughts @crawford?

kayrus: Rob, what do you mean?

kayrus:
Any news?

os/booting-on-vagrant.md
robszumski: I'm not sure if we've ever told anyone to stop a unit in cloud-config before.

kayrus: It doesn't harm

kayrus:
@robszumski any news?

os/booting-on-vagrant.md
crawford: Well, we are also turning on etcd2 and fleet, even though people don't necessarily need it either. Does it cause trouble if flannel is running?

os/booting-on-vagrant.md
crawford: You can just omit this line and it won't start by default.

os/booting-on-vagrant.md
kayrus: Lots of people want complete solution. I will just leave this string without "command: stop"

kayrus:
@crawford fixed

kayrus:
Have removed ruby code as it was already uncommented in vagrant PR:
https://github.com/coreos/coreos-vagrant/pull/265

crawford:
LGTM

PR #572, opened 2015-08-30
os/config-datetime: timedatectl output sync

joshix:
Update quoted `timedatectl` output to match that of recent systemd versions: `s/Ntp enabled/Network time on/`.

Update dates in quoted `timedatectl` output for freshness.


crawford:
Thanks! Can you squash your commits into a single commit and update the message (the one in the issue is perfect)?

PR #573, opened 2015-08-30
[2nd] os/config-datetime: timedatectl output sync

joshix:
Update quoted `timedatectl` output to match that of recent systemd versions: `s/Ntp enabled/Network time on/`.

Update dates in quoted `timedatectl` output for freshness.

Supersedes #572 to tidy commit history.


crawford:
LGTM. Thanks!

PR #574, opened 2015-08-31
os/mounting-storage: NFS: Refine language, add references.

joshix:
Clarify language.
Adopt and apply NFS convention of "export" rather than "share".
Add refs to external docs (see rpc.statd).


robszumski:
LGTM, thanks for making that a bit clearer.

PR #578, opened 2015-09-01
enterprise-registry: fix script url

robszumski:
fixes https://twitter.com/emeacham/status/638791348460605440

@coreoslinux https://coreos.com/products/enterprise-registry/docs/latest/mysql-container.html … needs updated; the link to the .sh file needs "mysql-container" removed from it.

cc: @josephschorr


josephschorr:
LGTM

Issue #575, opened 2015-08-31
docs on enabling TLS for docker


polvi:
https://groups.google.com/forum/#!topic/coreos-user/pGjRJakfLNA

would be helpful to add something specifically for this

kayrus:
@polvi can we close this issue?

ElijahCaine:
Closing due to inactivity.

PR #581, opened 2015-09-02
Added Docker TLS configuration doc

kayrus:
/cc @robszumski #575

I'm not sure which certificates tool should we use. But I suppose we have to promote our tools. If so, @yichengq should publish etcd-ca binaries for linux/darwin

### DIFF
--- a/os/customizing-docker.md
+++ b/os/customizing-docker.md
@@ -66,6 +66,122 @@ To keep access to the port local, replace the `ListenStream` configuration above
         ListenStream=127.0.0.1:2375
 ```
 
+## Enable the Remote API with TLS authentication
+
+Docker doesn't support HTTPS on systemd's `fd://` socket so it is impossible to configure socket activated docker service with TLS auth.
+
+Docker TLS configuration consists of two parts: keys creation and systemd drop-in configuration.
+
+### TLS keys creation
+
+Please follow the [instruction](generate-self-signed-certificates.md) to know how to create self-signed certificates and private keys. Then copy with following files into `/etc/docker` CoreOS' directory and fix their permissions:
+
+```sh
+scp ~/cfssl/{server.pem,server-key.pem,ca.pem} coreos.example.com:
+ssh core@coreos.example.com
+sudo mv {server.pem,server-key.pem,ca.pem} /etc/docker/
+sudo chown root:root /etc/docker/{server-key.pem,server.pem,ca.pem}
+sudo chmod 0600 /etc/docker/server-key.pem
+```
+
+On your local host copy certificates into `~/.docker`:
+
+```sh
+mkdir ~/.docker
+chmod 700 ~/.docker
+cd ~/.docker
+cp -p ~/cfssl/ca.pem ca.pem
+cp -p ~/cfssl/client.pem cert.pem
+cp -p ~/cfssl/client-key.pem key.pem
+```
+
+### Drop-in Configuration
+
+On remote CoreOS host create `/etc/systemd/system/docker.service.d/10-tls-verify.conf` [drop-in](using-systemd-drop-in-units.html) for systemd Docker service:
+
+```
+[Service]
+Environment="DOCKER_OPTS=-H=0.0.0.0:2376 --tlsverify --tlscacert=/etc/docker/ca.pem --tlscert=/etc/docker/server.pem --tlskey=/etc/docker/server-key.pem"
+```
+
+Reload systemd config files and restart docker service:
+
+```sh
+sudo systemctl daemon-reload
+sudo systemctl restart docker.service
+```
+
+Now you can access your Docker's API through TLS secured connection:
+
+```sh
+docker --tlsverify -H tcp://server:2376 images
+# or
+docker --tlsverify -H tcp://server.example.com:2376 images
+```
+
+If you've experienceed problems connection to remote Docker API using TLS connection, you can debug it with `curl`:
+
+```sh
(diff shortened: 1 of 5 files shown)


os/customizing-docker.md
```diff
@@ -66,6 +66,123 @@ To keep access to the port local, replace the `ListenStream` configuration above
...
+
+Now you can access your Docker's API through TLS secured connection:
+
+```sh
+docker --tlsverify -H tcp://%hostname%:2376 images
+# or
+docker --tlsverify -H tcp://example.com:2376 images
+```
+
+In addition you can export environmknt variables and use docker client without additional options:
```
robszumski: environment

kayrus:
Fixed environment

marineam:
We don't *have* to promote our own tools, in this case as long as it is a tool that's better than openssl :)

os/customizing-docker.md
```diff
@@ -66,6 +66,123 @@ To keep access to the port local, replace the `ListenStream` configuration above
...
+      content: |
+        ca.crt_content
+    - path: /etc/docker/ca.crt
+      permissions: 0644
+      content: |
+        %hostname%.crt_content
+    - path: /etc/docker/%hostname%.key.insecure
+      permissions: 0600
+      content: |
+        %hostname%.key.insecure_content
```
marineam: Including a full key and cert would be too much text but perhaps you could truncate one to a couple lines to make it clear what is supposed to be there. people tend to copy examples verbatim so making it more obvious that they do indeed need to edit the example would be good.

os/customizing-docker.md
```diff
@@ -66,6 +66,123 @@ To keep access to the port local, replace the `ListenStream` configuration above
...
+
+### TLS keys creation
+
+There are lots of instructions on how to create TLS keys. In this example we will show you how to create TLS keys with [etcd-ca](https://github.com/coreos/etcd-ca) tool. Please note that if you want to increase security you have to enter passphrase for the key. We assume that you will create keys on CoreOS host and then copy client's certificates to local machine.
+
+```sh
+# Initialize Certificate Authority
+./etcd-ca init --passphrase=""
+# Create certificate request for %hostname% docker server
+./etcd-ca new-cert --passphrase="" --domain="example.com" --ip="%private_ip%,%public_ip%" %hostname%
```
marineam: why the mixes of example.com and %hostname%, is the idea that %hostname% would be somehost.example.com? Also I generally prefer example values to substitution values, otherwise a user may think that the tool will perform the substitution. So `--ip="192.0.2.10,98.51.100.11"`

kayrus:
@yichengq , can you build release binaries for etcd-ca?

kayrus:
@robszumski , fixed examples

yichengq:
@kayrus We are moving from etcd-ca to cfssl now. cfssl is a general and actively maintained project to manage ca files. Check [this PR](https://github.com/coreos/etcd/pull/3321) for more details. /cc @gtank

kayrus:
Ok, it seems I have to change this PR again

gtank:
@kayrus Sorry about that. Feel free to ask me about any problems you run into.

kayrus:
Is there precompiled cfssl tool? Do we distribute it somehow? We can not tell people to download the sources and compile cfssl. Can we use [this][https://kuar.io/cfssl] url?

gtank:
Not sure, that's @kelseyhightower's project and I don't know if it's ever updated.

kayrus:
Could not find this link before.

kayrus:
@robszumski , i've added new doc. please review its text, as I suppose there could be problems.

@yichengq , can you review theoretical part? is it correct? should we provide TLDR part?

kayrus:
@yichengq can you also replace "g-loaded.eu" link with new one? https://github.com/coreos/etcd/blob/master/Documentation/security.md

robszumski:
Looks good to me other than one comment. Great job on that.

kayrus:
Fixed `***`

os/generate-self-signed-certificates.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Certificate types which are used inside CoreOS
+
+* **client certificate** is used to authenticate client by server. For example `etcdctl`, `etcd proxy`, `fleetctl` or `docker` clients.
+* **server certificate** is used by server and verified by client for server identity. For example `docker` server or `kube-apiserver`.
+* **client-server certificate** is used by etcd cluster members as they communicate with each other in both ways.
+
+### Configure CA options
+
+Now we can configure signing options inside `ca-config.json` config file. Default options conatin following preconfigured fields:
```
yichengq: contain

os/generate-self-signed-certificates.md
```diff
@@ -0,0 +1,301 @@
...
+* Please keep `ca-key.pem` file in safe, i.e. on your local host. This key allows to create any kind of certificates within your CA.
+* **\*.csr** files are not used and you can delete them.
+
+### Generate server certificate
+
+```
+cfssl print-defaults csr > server.json
+```
+
+Most important values for server certificate are **Common Name (CN)** and **hosts**. We have to substitute them, for example:
```
yichengq: may-be-related: I know some users want to have wild-card server certificate so they could use same server ceritificate in all servers. unsure about whether we need to support/guide that.

@gtank Thoughts?

yichengq:
LGTM.

kayrus:
Have fixed typo and added note about wildcard certificates.

robszumski:
I think this is ready to merge

marineam:
LGTM

PR #586, opened 2015-09-12
kubernetes: remove getting-started guide

robszumski:
This will now be synced from https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/getting-started.md


crawford:
LGTM

PR #588, opened 2015-09-17
Add SELinux documentation

mjg59:


os/selinux.md
```diff
@@ -0,0 +1,39 @@
...
+context, increasing isolation between containers even in the event of a
+container being compromised. CoreOS implements SELinux, but currently does
+not enforce SELinux protections by default.
+
+## Determining whether containers are compatible with SELinux policy
+
+To verify whether the current SELinux policy would inhibit your containers,
+enable SELinux logging by running the following commands as root:
+* rm /etc/audit/rules.d/80-selinux.rules
+* rm /etc/audit/rules.d/99-default.rules
```
marineam: Hm, the intent was to default to only logging SELinux stuff but I must have screwed up the rule somehow. We can document this as a work-around for now but we should get the rules fixed in next week's alpha.

marineam: Oh, also. those are symlinks and will get recreated on reboot iirc. replacing them with links to /dev/null or empty files would be needed to persistently remove them.

robszumski:
Any updates on this?

PR #595, opened 2015-09-21
os/sdk: include notes on dealing with the arm64 build

marineam:


philips:
great, lgtm. Perhaps we should make consistent "**ARM64 Note**" before each paragraph?

marineam:
Hm, giving it an extra "note" header or prefix seems like it would work best if the sections were wrapped in some sort of block quote formatting since the notes span several paragraphs w/ examples. Otherwise it still flows together and the note doesn't help that much for skimming. I don't think I've seen block quotes used on our site, do we have the right CSS and such to make that look ok?

robszumski:
Two headers sound good. One for "Building for amd64" and one for "Building for ARM". I'm suggesting that we specifically not include the 64 in ARM because its really hard to misread the headers since they both contain similar characters.

philips:
@robszumski @marineam it is technically aarch64

marineam:
yeah, arm64 because that's the Gentoo/Linux way or aarch64 because that's the official brand. 'ARM' is something else.

marineam:
but I don't really want to split amd64 out because there isn't any extra notes to add to amd64, it is just arm64 that had extra steps.

robszumski:
so what needs to happen here?

joshix:
Superseded and merged in #716

PR #598, opened 2015-09-22
Added VMware Workstation and Backdoor/Guestinfo notes

kayrus:
/cc @robszumski @crawford

### DIFF
--- a/os/booting-on-vmware.md
+++ b/os/booting-on-vmware.md
@@ -8,19 +8,37 @@ instructions as a starting point.
 
 ### Choosing a Channel
 
-CoreOS is released into alpha, beta, and stable channels. Releases to each
-channel serve as a release-candidate for the next channel. For example, a
-bug-free alpha release is promoted bit-for-bit to the beta channel.
+CoreOS is released into stable, alpha and beta channels. Releases to each channel serve as a release-candidate for the next channel. For example, a bug-free alpha release is promoted bit-for-bit to the beta channel. Read the [release notes][release notes] for specific features and bug fixes in each channel.
 
-The channel is selected based on the URLs below. Simply replace `stable` with
-`alpha` or `beta`. Read the [release notes][release notes] for specific
-features and bug fixes in each channel.
+<div id="vmware-images">
+  <ul class="nav nav-tabs">
+    <li class="active"><a href="#stable" data-toggle="tab">Stable Channel</a></li>
+    <li><a href="#beta" data-toggle="tab">Beta Channel</a></li>
+    <li><a href="#alpha" data-toggle="tab">Alpha Channel</a></li>
+  </ul>
+  <div class="tab-content coreos-docs-image-table">
+    <div class="tab-pane active" id="stable">
+      <div class="channel-info">
+        <p>Versions of CoreOS are battle-tested within the Beta and Alpha channels before being promoted. Current version is CoreOS {{site.stable-channel}}.</p>
+       </div>
+      <pre>curl -LO http://stable.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ova</pre>
+    </div>
+    <div class="tab-pane" id="alpha">
+      <div class="channel-info">
+        <p>The alpha channel closely tracks master and is released to frequently.  Current version is CoreOS {{site.alpha-channel}}.</p>
+      </div>
+      <pre>curl -LO http://alpha.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ova</pre>
+    </div>
+    <div class="tab-pane" id="beta">
+      <div class="channel-info">
+        <p>The beta channel consists of promoted alpha releases. Current version is CoreOS {{site.beta-channel}}.</p>
+      </div>
+      <pre>curl -LO http://beta.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ova</pre>
+    </div>
+  </div>
+</div>
 
-```sh
-curl -LO http://stable.release.core-os.net/amd64-usr/current/coreos_production_vmware_ova.ova
-```
-
-[release notes]: {{site.baseurl}}/releases
+[release notes]: https://coreos.com/releases/
 
 ### Booting with VMware Fusion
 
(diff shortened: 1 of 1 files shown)


os/booting-on-vmware.md
```diff
@@ -45,11 +63,20 @@ The last step uploads the files to your ESXi datastore and registers your VM.
...
+4. press "Retry" if VMware Workstation has raised "OVF specification" warning
+5. edit VM settings if necessary
+6. start up your CoreOS VM
+
 ## Cloud-Config
 
-Cloud-config can be specified by attaching a [config-drive][config-drive] with
-the filesystem label `config-2`. This is commonly done through whatever
-interface allows for attaching CD-ROMs or new drives.
+Cloud-config can be specified by attaching a [config-drive][config-drive] with the filesystem label `config-2`. This is commonly done through whatever interface allows for attaching CD-ROMs or new drives. In addition in CoreOS 801.0.0 and greater you can use [VMware GuestInfo][vmware backdoor].
```
robszumski: Based on the trouble that everyone has with the config-drive, should we just drop this altogether and rely on the guest info stuff?

os/booting-on-vmware.md
```diff
@@ -45,11 +63,20 @@ The last step uploads the files to your ESXi datastore and registers your VM.
 You can now tweak the VM settings, like memory and virtual cores, then power it
 on. These instructions were tested to deploy to an ESXi 5.5 host.
 
+### Booting with VMware Workstation 12
+
+Run VMware Workstation GUI:
```
robszumski: Would screenshots help here? My familiarity with these tools is lacking so if people can easily understand this I'm fine with no screenshots.

os/booting-on-vmware.md
kayrus: It is possible to combine configdrive and vmware backdoor. As it is not easy to provide cloud-config through vmware (you have to encode it), it is more conveniently to use config drive.

os/booting-on-vmware.md
kayrus: I don't have these products. if someone provide screenshots - we can add them. But we have to note that we have to update them from time to time on every GUI change.

os/booting-on-vmware.md
robszumski: I was going to send this to a user, but I'm still not clear how you use the backdoor. This either needs to be covered here or on the backdoor doc. Where do I put those variables? How can I use them to do something useful? Can I see an example cloud-config or scenario?

kayrus: Examples are in coreos-cloudinit docs under the link beyond. I don't know where exactly to put detailed info. Of course we can put full info here, but i'm not sure if it is correct place for that info.

kayrus:
@robszumski Moved info from https://github.com/coreos/coreos-cloudinit/pull/393

robszumski:
cc @joshix who is going to do a fresh review of this doc in the nearish future

kayrus:
@joshix any news?

kayrus:
@robszumski @joshix fixed links

os/booting-on-vmware.md
```diff
@@ -58,11 +85,104 @@ retained, although CoreOS supports config-drive on all platforms.
...
+### Defining Cloud-Config in GuestInfo
+
+You have to encode your regular Cloud-Config with `base64` tool and set output into `coreos.config.data` variable.
+
+|  encoding   |                      command                      |
+|:------------|:--------------------------------------------------|
+| base64      | `base64 -w0 /path/to/user_data && echo`           |
+| gzip+base64 | `gzip -c /path/to/user_data | base64 -w0 && echo` |
+
+Or simply set URL in `coreos.config.url` variable.
```
robszumski: To avoid having to base64, you can retrieve your Cloud-Config from a URL with the `coreos.config.url` variable.

question: is this true? or do you need to base64 encode that too?
question: do these variables need `guestinfo.`?

os/booting-on-vmware.md
```diff
@@ -58,11 +85,104 @@ retained, although CoreOS supports config-drive on all platforms.
...
+
+[Address]
+Address=192.168.178.97/24
+
+[Route]
+Destination=0.0.0.0/0
+Gateway=192.168.178.1
+```
+
+This unit file will configure the matching network interface. If you have set `coreos.config.data` or `coreos.config.url` variables, `coreos-cloudinit` will try to apply these Cloud Configs on VM and substitute `$private_ipv4` and `$public_ipv4` variables if you've configured network interfaces' roles using `interface.<n>.role` guest variable.
```
robszumski: do these need `guestinfo.`?

os/booting-on-vmware.md
```diff
@@ -58,11 +85,104 @@ retained, although CoreOS supports config-drive on all platforms.
...
+
+### Defining Cloud-Config in GuestInfo
+
+You have to encode your regular Cloud-Config with `base64` tool and set output into `coreos.config.data` variable.
```
robszumski: `guestinfo.`?

robszumski: There are two supported encoding types:

os/booting-on-vmware.md
```diff
@@ -58,11 +85,104 @@ retained, although CoreOS supports config-drive on all platforms.
...
+
+|  encoding   |                      command                      |
```
robszumski: title case:

"Encoding" and "Command"

robszumski:
Added a few small comments

os/booting-on-vmware.md
robszumski: According to @crawford this is true, it can't be base64 encoded.

robszumski: This brings up another question, does this support https? Should we add an example below?

crawford: It should be obvious from the variable names that `coreos.config.data.encoding` applies to `coreos.config.data` and not `coreos.config.url`.

robszumski: It wasn't obvious to me. It's worth calling out.

robszumski: actually, we don't state that you need to set both. the example does include this however.

os/booting-on-vmware.md
robszumski: updated:

There are two supported encoding types, which can be configured with the `guestinfo.coreos.config.data.encoding` variable:

kayrus:
Moved all links into the footer, and fixed your comments

os/booting-on-vmware.md
```diff
@@ -58,11 +85,99 @@ retained, although CoreOS supports config-drive on all platforms.
 For more information on customization that can be done with cloud-config, head
 on over to the [cloud-config guide][cloud-config guide].
 
-Note: The `$private_ipv4` and `$public_ipv4` substitution variables referenced
-in other documents are *not* supported on VMware.
+The `$private_ipv4` and `$public_ipv4` substitution variables are fully supported in cloud-config on VMware but only in CoreOS release 801.0.0 and greater. These releases has updated [coreos-cloudinit][coreos-cloudinit] v1.6.0 which allows you to configure network using [VMware GuestInfo options][vmware backdoor] and define `public` and `private` roles for ethernet interfaces.
+
+## VMware Backdoor
```
joshix: VMWare Guestinfo Interface

(We are trying to eliminate use of the phrase "backdoor". See https://github.com/coreos/coreos-cloudinit/pull/404

PR #609, opened 2015-10-02
Update registry-authentication.md

krish7919:
The '<username>' & '<password>' characters are not rendered on the browser correctly.
Solution: Escaping them using backquotes.


robszumski:
Looks good to me, thanks!

PR #618, opened 2015-10-12
*: add our golang style and development guidelines

jonboulle:
We were previously hosting those internally, but as the vast majority of
CoreOS projects are developed in the open, we should share them and have
them publicly referenceable for anyone working on CoreOS projects.


philips:
lgtm

PR #624, opened 2015-10-14
Add info on changing sshd listen port after install.

omkensey:


os/customizing-sshd.md
```diff
@@ -1,6 +1,8 @@
-# Customizing the SSH Daemon
+ustomizing the SSH Daemon
```
joshix: Customizing (typo)

os/customizing-sshd.md
```diff
@@ -43,6 +45,99 @@ coreos:
       Accept=yes
 ```
```
joshix: It would be nice to explain the result of following the example here. E.g.,  "Sshd will now be activated by requests to port #2222, rather than 22", or something like that.

os/customizing-sshd.md
```diff
@@ -26,7 +28,7 @@ write_files:
       ChallengeResponseAuthentication no
 ```
 
-## Changing the sshd Port
+### Changing the sshd Port
 
 CoreOS ships with socket-activated SSH by default. The configuration for this can be found at `/usr/lib/systemd/system/sshd.socket`. We're going to override this in the cloud-config provided at boot:
```
joshix: Does "override this" mean "override the configured listen port"?

joshix:
LGTM with amendments suggested in line comments. Kick it back after looking there and I'll merge. Definitely useful and informative -- we get that question about disabling password-auth with some frequency, IIRC.

PR #633, opened 2015-10-21
Begin style guide: Document conventions

joshix:
~~Add a `style/` directory.~~
Add STYLE.md guide to conventions and standards. Beginning with headings.

Addresses https://github.com/coreos-inc/coreos-pages/issues/324

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,32 @@
 # CoreOS Documentation
 
-These are publicly accessible versions of the documentation on [coreos.com](http://coreos.com/docs/).
+This repository contains the Markdown source of the [CoreOS documentation][coreos-docs].
 
-Check out the [help-wanted tag](https://github.com/coreos/docs/issues?q=is%3Aopen+label%3Ahelp-wanted) if you're not sure what to help out with.
+## Contributing
 
-1. Fork this repo
-2. Send a Pull Request
-3. We will get your fix up on [coreos.com](http://coreos.com/docs/)
+The CoreOS documentation is released under the [Apache 2.0 License][asl], and we welcome contributions. Check out the [help-wanted tag][help-wanted] in this project's Issues list for good places to start participating.
 
-Thanks for your contributions and fixes!
+Submit fixes and additions in the form of `git` *Pull Requests* (PRs). The general process is the typical git fork-branch-PR-review-merge cycle:
 
-## Translations
+1. Fork this repository into your git workspace and/or GitHub account
+2. Make changes in a topic branch or your fork's `master`
+3. Send a Pull Request from the branch containing the changes
+4. Maintainers will review the PR and either merge it or make comments
 
-We accept patches for documentation translation. Please send the
-documents as a pull request and follow two guidelines:
+### Style guide
 
-1. Name the files identically to the originals but put them in a
-   directory that matches the language locale. For example: JA\_JP,
-   ZH\_CN or KO\_KN.
+The [style guide][style] prescribes the conventions of formatting and English style preferred in CoreOS project documentation.
 
-2. Add an explanation about the translated document to the top of the
-   file: "These documents were localized into Esperanto by Community
-   Member <person@example.com> and last updated on 2014-04-04. If you
-   find inaccuracies or problems please file an issue on GitHub."
+### Translations
 
-Thank you for your contributions.
+We happily accept accurate translations. Please send the documents as a pull request and follow two guidelines:
+
+1. Name the files identically to the originals, but put them beneath a directory named for the translation's `gettext` locale. For example: `JA_JP/doc.md`, `ZH_CN/doc.md,` or `KO_KN/doc.md`.
+
+2. Add an explanation about the translated document to the top of the file: "These documents were translated into Esperanto by Community Member <person@example.com> and last updated on 2015-12-01. If you find inaccuracies or problems please file an issue on GitHub."
+
+
+[asl]: LICENSE
+[coreos-docs]: https://coreos.com/docs/
+[help-wanted]: https://github.com/coreos/docs/issues?q=is%3Aopen+label%3Ahelp-wanted
+[style]: STYLE.md "CoreOS Documentation Style and Formatting"
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,187 @@
+# Documentation Style and Formatting
+
+## English style
+
+Write short sentences. Organize concepts in paragraphs. Prefer lists to tables and paragraphs to lists. Write in the active voice. Avoid jargon beyond the requirements of subject and audience.
+
+### Eschew you
+
+You write unambiguous documentation, so you avoid the second person. Avoiding personal pronouns in general helps produce the imperative impersonal tone desired for documentation. Don't reboot your system or have the user reboot their system. Reboot the system.
+
+### Generalities
+
+There are a few other common ways to write or not write things:
+
+* Expand acronyms on their introduction in a document, with the short form following in parentheses: Trusted Platform Module (TPM)
+* Terms of art that are not commands or other literal text should often be italicized on their first appearance in a document: *Kubernetes* is a good example
+* The hyphen is overused and most English compounds do not require it
+
+## Source formatting
+
+CoreOS documentation is written in [Markdown][mdhome], a simple way to annotate text to indicate presentation typesetting. Markdown source is intended to be a plain text human-readable version of the document, even before conversion to HTML for the browser or other display.
+
+### Source file naming and encoding
+
+Write Markdown source in UTF-encoded plain text files, named with a reasonable, lower case short form of the document's title, and suffixed with `.md`. Prefer hyphens to underscores in file names with two or more words. For example, instructions for DNS configuration are written to a file named [`configuring-dns.md`][configuring-dns].
+
+### Line wrapping considered harmful
+
+Don't wrap long lines of text with manual newlines. Line wrapping churns prose documents, because lines not actually edited will nevertheless change when a paragraph is edited and rewrapped.
+
+### One sentence per line deprecated
+
+Writing Markdown source with a newline between every sentence is acceptable to most compilers and can ease change review. However, this format makes the document less readable in source form. Do not add a line break between sentences. Write natural English paragraphs, separated by a single blank line.
+
+### Preferred markdown symbols
+
+Markdown defines two or more ways to declare some document structures. This documentation prefers these Markdown symbols among their alternatives:
+
+* Headings are denoted in Markdown's ATX style, with hash character(s): `#`. See [*Headings*][headings], below.
+* Bulleted lists, like this one, are denoted with the asterisk (`*`), rather than the hyphen.
+* Hyperlink URLs are given in the reference style (`[hyperlinked text][label]`), rather than inline. Hyperlink labels are defined in one list at the end of the document. Relative links are preferred to absolute links. See [*Hyperlink Considerations*][hyperlink-considerations], below.
+* *Italic text* is wrapped with a pair of single asterisks: `*Italics*`; **Bold** with a double pair: `**Bold**`.
+* `Monospace` is indicated between a pair of backticks. This distinguishes literal strings like command names, file paths, or values, e.g., `/bin/markdown`. See [*Command Line Grammar*][command-line-grammar], below.
+* Longer code blocks or file contents are *fenced*: Set off on new lines between pairs of three backticks, rather than indented. A presentation hint specifying the block's language can be given immediately after the opening three backticks, e.g., ````yaml`.
+
+## Headings
+
+By convention, the level one heading, denoted in Markdown by a single hash character (`#`), is the document's title. This document's title is *Documentation Style and Formatting*. The document title, and only the document title, is typed in *Title Case*.
+
+### Heading style
+
+Each heading is both preceded and followed by a newline. A space separates the Markdown symbols from the heading text. Headings other than the single level one title are typed in *Sentence case*, capitalizing the first letter of the first word, but other words only as they would be capitalized if appearing in the middle of a sentence.
+
+### Heading semantics and the sidebar outline
+
+Section headings expose the document's logical structure with a notation of incrementing hash marks (`#[#][...]`) for increasingly nested levels of a hierarchy. With the level one heading devoted to the document title, the second-level headings represent the document's primary concepts.
+
+The site deployment process inspects a document's headings to derive the thumb index outlines seen in the right sidebar of [documentation viewed at CoreOS.com][coreos-docs].
+
+#### Example: This document's source
+
+The abridged skeletal markdown source for this document's headings:
+
+```
+# Documentation Style and Formatting
+
+## English style
+
+### Eschew you
+
+[...]
+
+## Headings
+
+### Heading style
+
+### Heading semantics and the sidebar outline
+
+[...]
+
+## Command line grammar
+
+### Terms
+
+[...]
+
+## File name extension conventions
+
+```
+
+### Example: The "average" document
+
+Most documents have a single `h1` (`#`) heading matching the title, two to five `h2` (`##`) headings representing the topic's primary concepts, and one or two `h3` (`###`) and `h4` (`####`) headings organizing details beneath each `h2`.
+
+If a document proves a great deal longer or more structurally complex than those simplistic rules of thumb, there should be a good reason.
+
+## Hyperlink considerations
+
+### Naming
+
+Name hyperlinks carefully to give them maximum context. For example, note that certain information is in the [style guide][style], rather than just pointing lazily to the style guide [here][style]. The link text "here" gives almost no information about its target. It is helpful to [write a clear sentence][eos] first, then bracket the choice words within to declare them a hyperlink.
+
+### Marking down the link
+
+As mentioned above, the reference style of Markdown hyperlinking is preferred to the inline. Hyperlinks are marked with two pairs of square brackets, the first enclosing the hyperlinked text, the second enclosing a label for the link. Labels are in turn associated with a target URL in a list of declarations at the end of the document. Each label declaration consists of a line beginning with the bracket-enclosed label, a colon, and the target URL (the `href` in HTML). The target URL may optionally be followed by a link title in double quotes. The list of link label declarations should be sorted alphabetically.
+
+#### Example: Reference-style hyperlinking
+
+```markdown
+The reference style of [Markdown hyperlinks][mdlinks] allows for easier
+reading of source and formalizes the declaration of links.
+
+Another paragraph may reference the [project introduction][readme],
+which link will likewise have its label defined at the document's foot.
+
+[mdlinks]: http://daringfireball.net/projects/markdown/syntax#link "Markdown link syntax"
+[readme]: README.md
+```
+#### Relative URLs preferred
+
+Using relative URLs where possible helps portability among multiple presentation targets, as they remain valid even as the site root moves. Absolute linking is obviously necessary for resources external to the document's repository and/or the coreos.com domain.
+
+For example, there are two ways to refer to the [CoreOS quick start guide][quickstart]'s location. The preferred way is a relative link from the current file's path to the target, which from this document is `os/quickstart.md`. An absolute link to the complete URL is less flexible, and more verbose: `https://github.com/coreos/docs/blob/master/os/quickstart.md`.
+
+#### Hyperlink deployment automation
+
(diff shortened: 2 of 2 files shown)


style/headings.md
```diff
@@ -0,0 +1,30 @@
+# Headings in CoreOS Documentation
+`# Headings in CoreOS Documentation`
```
robszumski: I wasn't sure what to make of this at first. Maybe we have an example that uses the headings in the doc as an example and it looks something like:

For example, this document's headings are structured in code as:

```
# Headings in CoreOS Documentation

## Documentation Format

### Subheadings

#### Subheadings and the Sidebar Outline

### Example: *The "Average" Document*
```

style/headings.md
```diff
@@ -0,0 +1,30 @@
...
+
+This document is a guide to creating the headings that outline a document's structure. By convention, the level-one heading, denoted in markdown by a single hash character (`#`), is the document's title. This document's title is *Headings in CoreOS Documentation*.
+
+In this document, each heading is followed by the literal string representing it in the markdown source. In all documentation, `monospace` represents literal strings in example commands or source, e.g., `# Headings in CoreOS Documentation`, as above.
+
+## Documentation format
```
robszumski: title case

robszumski:
Should we go ahead and make the whole styleguide one document, ie `styleguide.md` and keep it in the root dir, like the license, etc?

As much as I hate to suggest this, maybe do all caps like the rest, ie `STYLEGUIDE.md`

robszumski:
We can always break it apart later if needed

style/headings.md
```diff
@@ -0,0 +1,30 @@
...
+
+Subsection headings expose the document's logical structure with a notation of incrementing hash marks (`#[#][...]`) for increasingly nested levels of a hierarchy.
+
+#### Subheadings and the Sidebar Outline
+`#### Subheadings and the Sidebar Outline`
+
+The site deployment process inspects a document's headings to derive the thumb-index outlines seen in the right sidebar of documentation viewed at [CoreOS.com](https://coreos.com/docs/).
+
+### Example: *The "Average" Document*
+`### Example: "The Average Document"`
```
jonboulle: This is inconsistent with the preceding line

jonboulle:
Perhaps not in this PR but would love to address the wrapping issue raised in coreos/rkt#1648. Generally speaking, we've tried to stick to a "one sentence per line" policy in markdown documentation because it makes it much easier to review diffs on GitHub.

robszumski:
@jonboulle I break this rule all over the place, so we should try to figure out a rule around that.

jonboulle:
dragging in @philips ...

joshix:
@robszumski Re: file names/directory anticipation: I'm so much in the undecided column that I started on two previous drafts with the file name `styleguide.md`, then `STYLEGUIDE.md` in observance of observed local custom. However I moved it into a directory because then each file beneath `style/` may have a shorter name and this particular iteration can omit an introduction explaining it is not in fact the Style Guide, etc. This to explain my thinking, rather than argue with urgency for either side.

I had thought about some more-developed later version living in a file named `coreos/docs/CONTRIBUTING.md`, superseding this PR, following the model of, and linked to from within, existing code contribution guidelines like https://github.com/coreos/etcd/blob/master/CONTRIBUTING.md.

joshix:
@jonboulle So the source for markdown would look like:

```
# Markdown line breaks

This is a sample document showing the proposed one-sentence-per-line formatting convention.
This second sentence continues the same paragraph.

This third sentence begins a second paragraph.

## Second heading

This sentence is short.
This sentence is longer.
This sentence is longest.

This sentence is lonely.
```
Renders as:

# Markdown line breaks

This is a sample document showing the proposed one-sentence-per-line formatting convention.
This second sentence continues the same paragraph.

This third sentence begins a second paragraph.

## Second heading

This sentence is short.
This sentence is longer.
This sentence is longest.

This sentence is lonely.

####

Other renderers ignore the single-LF and join lines so separated into paragraphs, but the github renderer seems to obey rather than ignore them.

jonboulle:
I'm confused. We do this successfully elsewhere. Let me look when in front
of keyboard again.

joshix:
I think the github md interpreter gives different output here in a comments chain vs on displaying a source file, e.g., coreos/docs/README.md. That file when experimentally adjusted toward one-sentence-per-file-line displays as expected, with single LF usually just consumed and ignored. See https://github.com/joshix/docs/commit/81debe73dfd3225c8d65acf40bcd5ad32a20ccc5

## One-sentence-per-line:
### Pro:
* Definitely easier to diff and review.

### Con:
* More tedious to write
* More tedious to read as source and/or in terminal
* Does not address 80-char width also discussed in related issue(s)

jonboulle:
I'm still in favour of optimising for diff reviews, and am not really concerned about the tediousness of writing it.
Crazy compromise: 80-char width but new sentence always means newline?!

crawford:
I don't really like the idea of imposing this restriction on all teams just because some teams use poor tooling to review diffs (read GitHub (which still blows my mind)). I'm in favor of an 80-char limit (without the proposed line break) since I get work done in a terminal and that is where the docs need to be readable.

jonboulle:
@joshix care to make an executive decision on this one?

joshix:
Sure. Let me hang the exit sign on the bikeshed door:

1. One-sentence-per-line: Nice idea. But formatting concerns begin to tread on the ideas the reader is trying to understand, the writer is trying to communicate. The text becomes fussy and stops looking like paragraphs when read in markdown source form. The value added in reviewing is, to me, minimal. Write natural english sentences and paragraphs, not one-sentence-per-line.

2. Character per line limit: Obey projects' 80-char per line rule. URLs and other literal formations can break this rule for a line or two. Use common sense, but generally wrap your lines at around char 80. 
Markdown compiler is the source of truth.

I intend to write as much into the CONTRIBUTING guide and update this PR, unless anyone thinks I should be set on fire instead.

Cf. https://github.com/google/styleguide/blob/gh-pages/docguide/style.md#character-line-limit

jonboulle:
SGTM, probably best to get an ack from @philips as AIUI he's the initiator
of the one-sentence-per-line style

philips:
I still like one sentence per line and it works in nearly all editors. The 80 char rule gets you into doc churn hell because everyone's editor is going to do something different whether that be wrapping to 79, wrapping words with dashes, etc.

joshix:
@philips Since we don't agree, let me spell out my reasoning and you can make the ruling. I thought about the points you raise, but I came down to two points: 1) as @crawford alludes, part of the intent of Markdown is plain-text readability. The source document is supposed to be a fair rough representation of the document as rendered. We write the kind of docs with markdown that one would expect to read in a terminal. The one sentence-per-line rule makes this pretty terrible. Nothing looks like a paragraph. I presumed (not knowing the history) that markdown was chosen for these conventional reasons and therefore those goals were worth preserving.

Having thought it out that way, I compared with some style guides for large repos using Markdown. Google's rules pretty much encapsulate what I was aiming for. Maybe that's confirmation bias. https://github.com/google/styleguide/blob/gh-pages/docguide/style.md#character-line-limit

In order, I would happily concede on the proposed 80-chars per line. I truly think the intent of markdown is writing natural english paragraphs, and I kind of hate `humanfmt` in general.

I would be forced to concede, but like it much less, if we had to adopt one-sentence-per-line all over.

joshix:
Note: spec unordered list style (`* Item`)

krnowak:
Just my X cents here:

I think that introducing one style over all CoreOS projects by asking people for their opinions is a mistake. The more people you ask the more strong and mutually exclusive opinions you get.

Right now there is a contention over "one sentence per line" (@jonboulle, @philips) vs "limit to 80 chars" (@crawford) paragraph style. Same with "New York Time" (@joshix) and "Sentence without period" (@jonboulle) heading style. I'd allow both styles for paragraphs and heading, but with a caveat that each project have to decide which style it will use, otherwise it's New York Times with 80 chars limit.

jonboulle:
I should clarify that my preference is not strictly sentence-per-line, but rather avoiding line length limits. I'm +1 on the proposed paragraph-per-line.

joshix:
YAML files: `secret.yaml` vs `secret.yml`

Set style for this. Before deciding, note that upstream k8s tends to say `*.yaml`.

joshix:
@robszumski @meghanschofield @philips I think we could land this as a delightful holiday gift for one and all. Would enjoy your parting thoughts on same.

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+Most documents have a single `h1`-level (`#`) heading matching the title, two to five `h2`-level (`##`) headings representing the topic's primary concepts, and one or two `h3` (`###`) and `h4` (`####`) headings organizing details beneath each `h2`.
+
+If a document proves a great deal longer or more structurally complex than those simplistic rules of thumb, there should be a good reason.
+
+## Hyperlink Considerations
+
+### Naming
+
+Name hyperlinks carefully to give them maximum context. For example, note that certain information is in the [style guide](STYLE.md), rather than just pointing lazily to the style guide [here](STYLE.md). The link text "here" gives almost no information about its target. It is helpful to [write a clear sentence](https://faculty.washington.edu/heagerty/Courses/b572/public/StrunkWhite.pdf) first, then bracket the choice words within to declare them a hyperlink.
```
robszumski: yes yes yes. I hate saying "these are available [here]()"

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+Writing Markdown source with a newline between every sentence is acceptable to most compilers and can ease change review. However, this format makes the document less readable in source form. Do not add a line break between sentences. Write natural English paragraphs, separated by two newlines.
+
+### Preferred Markdown Symbols
+
+Markdown defines two or more ways to declare some document structures. This documentation prefers these Markdown symbols among their alternatives:
+
+* Headings are denoted in Markdown's ATX style, with hash character(s): `#`. See [*Headings*](#headings), below.
+* Bulleted lists, like this one, are denoted with the asterisk (`*`), rather than the hyphen.
+* Hyperlink URLs are given inline (`[hyperlinked text](target.md)`), rather than reference-style. Relative links are preferred to absolute links. See [*Hyperlink Considerations*](#hyperlink-considerations), below.
```
robszumski: We want reference style sometimes, right? For times when the same link is used more than once is a possible rule for that?

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+*Commands* *invoke* or *execute* programs. Commands *take* *arguments* and *accept* *options*, which themselves may be *set* to *values*.
+
+### Example: Documenting `echo(1)`
+
+In this simple command line:
+
+```sh
+$ echo -n Example
+Example$
```
robszumski: I get what this is showing, but seems more confusing then just leaving it off. When it is every absolutely necessary to get this 100% correct?

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+`echo` is the command, and `Example` is the argument. The option `-n` suppresses the terminating newline usually emitted by `echo`. A single-letter binary option like `-n` is sometimes called a *flag*. The `echo(1)` command prints its argument on the standard output, and a good shell excerpt often includes the expected output of commands, as shown here. The shell prompt character `$` distinguishes input from output.
+
+### Example: Documenting Subcommands
```
robszumski: Do we want to cover long commands here?

Do this:

```
$ docker run \
  -d \
  -p 80:80 \
  registry.example.com:5000/apache \
  /usr/sbin/apache2ctl -D FOREGROUND
```

instead of:

```
$ docker run -d -p 80:80 registry.example.com:5000/apache /usr/sbin/apache2ctl -D FOREGROUND
```

robszumski: Open question: do you indent the lines or not?

robszumski:
Added a few comments, but overall looks good.

STYLE.md
joshix: The truncated output line (133) in particular, or the entire example?

robszumski: Yeah just the `$` on 133

STYLE.md
joshix: I think we want a simple single rule more than occasional exceptional convenience, but I also don't plan a mass rewrite to kick off tomorrow. How certain are you that we need the exception?

jonboulle: 2€¢: after writing a pile o' markdown over the last couple of years, I've started trending towards preferring reference style; completely anecdotally, a slight majority of URLs end up making the source pretty unfriendly to read when referenced inline.

README.md
```diff
@@ -1,27 +1,26 @@
...
-1. Fork this repo
-2. Send a Pull Request
-3. We will get your fix up on [coreos.com](http://coreos.com/docs/)
+The CoreOS documentation is released under the [Apache 2.0 License](LICENSE), and we welcome contributions. Check out the [help-wanted tag](https://github.com/coreos/docs/issues?q=is%3Aopen+label%3Ahelp-wanted) in this project's Issues list for good places to start participating.
 
-Thanks for your contributions and fixes!
+Submit fixes and additions in the form of `git` *Pull Requests* (PRs). The general process is the typical git fork-branch-PR-review-merge cycle:
 
-## Translations
+1. Fork this repository into your git workspace and/or github account
```
jonboulle: GitHub

README.md
```diff
@@ -1,27 +1,26 @@
...
+2. Make changes in a topic branch off your fork's `master`
```
jonboulle: This is unnecessarily prescriptive; I submit pull requests from my personal `master` branch all the time

STYLE.md
```diff
@@ -0,0 +1,147 @@
+# Documentation Style and Formatting
+
+## English Style
+
+Write short sentences. Organize concepts in paragraphs. Prefer lists to tables and paragraphs to lists. Write in the active voice. Avoid jargon beyond the requirements of subject and audience.
+
+### Eschew You
```
jonboulle: https://www.youtube.com/watch?v=ya2Sx8xmMpc

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+You write unambiguous documentation, so you avoid the second person. Avoiding personal pronouns in general helps produce the imperative impersonal tone desired for documentation. Don't reboot your system or have the user reboot their system. Reboot the system.
+
+### Generalities
+
+There are a few other common ways to write or not write things:
+
+* Expand acronyms on their introduction in a document, with the short form following in parentheses: Trusted Platform Module (TPM)
+* Terms of art that are not commands or other literal text should often be italicized on their first appearance in a document: *Kubernetes* is a good example
+* The hyphen is overused and most English compounds do not require it
```
jonboulle: over-used, c'mon

jonboulle: com-pounds

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+Write Markdown source in UTF-encoded plain text files, named with a reasonable, lower-case short form of the document's title, and suffixed with `.md`. Prefer hyphens to underscores in multi-word file names. For example, instructions for DNS configuration are written to a file named [`configuring-dns.md`](os/configuring-dns.md).
+
+### Line Wrapping Considered Harmful
+
+Don't wrap long lines of text with manual newlines. Line wrapping churns prose documents, because lines not actually edited will nevertheless change when a paragraph is edited and re-wrapped.
+
+### One Sentence per Line Deprecated
+
+Writing Markdown source with a newline between every sentence is acceptable to most compilers and can ease change review. However, this format makes the document less readable in source form. Do not add a line break between sentences. Write natural English paragraphs, separated by two newlines.
```
jonboulle: "two newlines" is arguably ambiguous, perhaps "one blank line"?

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+### Preferred Markdown Symbols
+
+Markdown defines two or more ways to declare some document structures. This documentation prefers these Markdown symbols among their alternatives:
+
+* Headings are denoted in Markdown's ATX style, with hash character(s): `#`. See [*Headings*](#headings), below.
+* Bulleted lists, like this one, are denoted with the asterisk (`*`), rather than the hyphen.
+* Hyperlink URLs are given inline (`[hyperlinked text](target.md)`), rather than reference-style. Relative links are preferred to absolute links. See [*Hyperlink Considerations*](#hyperlink-considerations), below.
+* *Italic text* is wrapped with a pair of single asterisks: `*Italics*`; **Bold** with a double pair: `**Bold**`.
+* `Monospace` is indicated between a pair of backticks. This distinguishes literal strings like command names, file paths, or values, e.g., `/bin/markdown`. See [*Command Line Grammar*](#command-line-grammar), below.
+* Longer code blocks or file contents are set off on new lines between pairs of three backticks, rather than with indentation. A presentation hint of the block's language can be given immediately after the opening three backticks.
```
jonboulle: no example for this one?

STYLE.md
jonboulle: hm, [here](https://github.com/coreos/docs/pull/633/files#diff-32430379d7ba797ddf827c29a4eac4c1R52) you define a "new line" (c.f. "newline")

STYLE.md
jonboulle: **shakes fist**

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Hyperlink Considerations
+
+### Naming
+
+Name hyperlinks carefully to give them maximum context. For example, note that certain information is in the [style guide](STYLE.md), rather than just pointing lazily to the style guide [here](STYLE.md). The link text "here" gives almost no information about its target. It is helpful to [write a clear sentence](https://faculty.washington.edu/heagerty/Courses/b572/public/StrunkWhite.pdf) first, then bracket the choice words within to declare them a hyperlink.
+
+### Marking Down the Link
+
+As mentioned above, the inline style of hyperlinking is preferred to other ways of writing hyperlinks in Markdown. Using relative URLs where possible helps shorten these inline hyperlink declarations. In addition, relative links remain valid even as the site root moves. Absolute linking is obviously necessary for resources external to the document's repository and/or the cores.com domain.
```
jonboulle: cores -> coreos

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+
+There are two ways to refer to the [quick start guide](os/quickstart.md):
+
+1. The preferred way is a relative link from the current file's path to the target: `[quick start guide](os/quickstart.md)`
+2. An absolute link is less flexible: `[quick start guide](https://github.com/coreos/docs/blob/master/os/quickstart.md)`
+
+
+#### Hyperlink Deployment Automation
+
+CoreOS documents have two major publication targets: the [coreos.com documentation library](https://coreos.com/docs/), and github's in-repo Markdown presentation. The deployment scripts used to build the CoreOS site handle some of the wrinkles arising between the two targets. These scripts expect links to other CoreOS project documentation to refer to the Markdown source, that is, to end with the `.md` file extension. The deployment scripts rewrite hyperlinks to replace that extension with `.html` for presentation. This allows the links to be valid in either context. External links are not rewritten.
```
jonboulle: GitHub

jonboulle: `Markdown source, that is, to end`
-->
` Markdown source; that is, to end`

#teamsemicolon

STYLE.md
jonboulle: yeah, imho `-n` is a confusing flag to choose.

STYLE.md
jonboulle: +1 would be good to have a referenceable answer to this one

STYLE.md
crawford: I personally prefer reference-style as well. Mainly because I find it more readable. Otherwise, I have to train my eyes to recognize and skip over URLs.

STYLE.md
```diff
@@ -0,0 +1,147 @@
...
+### Example: Documenting Subcommands
+
+Some command lines are more complex. Many commands operate through a set of *subcommands*. `rkt` and several other relevant programs follow this pattern.
+
+```sh
+$ rkt run --debug example.aci
+[...]
+```
+
+In this case the argument to `rkt` is a subcommand, `run`. `run` in turn accepts the `--debug` option to modify how it executes the ACI image specified by its own argument, `example.aci`
```
crawford: In this case the argument to `rkt`, `run`, is a subcommand.

STYLE.md
joshix: `-n` is the only option to `echo(1)`!

Replacing with [`cat -v`](http://harmful.cat-v.org/cat-v/)...

You two are correct in your suggestions here but I am still enjoying a good chuckle. I'll make this example a little less fiddly.

STYLE.md
joshix: Good call on long cmd lines. Will add example.

joshix: @robszumski I don't indent when I break cmd lines; it can interfere with copy-paste into finicky terminals (admittedly not often), and I think it looks odd when reproduced in the display code block.

I write markdown as lazily as I can figure out how to get away with. I also avoid the wrap-and-indent tricks beneath markdown lists. You end up with lots of false diffs on every changeset where the newlines are the only thing moving around.

STYLE.md
joshix: +1 Ralph Wiggam

PR #635, opened 2015-10-22
update-window.sh should handle both etcd and etcd2 installation.

xueshanf:
Right now the update-window.sh won't work properly on etcd2 installation - it would just randomly reboot etcd2 members without reboot locking.

### DIFF
--- a/os/update-strategies.md
+++ b/os/update-strategies.md
@@ -148,24 +148,40 @@ This script should be installed in a location to match the script path used in t
 
 ```yaml
 #!/bin/bash
+# If the system is a member of an etcd cluster, uses locksmithctl to reboot. 
+# Otherwise, it randomly delays the reboot.
 
-# If etcd is active, this uses locksmith. Otherwise, it randomly delays. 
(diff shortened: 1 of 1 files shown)


os/update-strategies.md
```diff
@@ -148,24 +148,40 @@ This script should be installed in a location to match the script path used in t
 
 ```yaml
```
joshix:
```sh

joshix:
Note: @kayrus is working on a different approach to this concept.

crawford:
Locksmith 0.3.4 has support for reboot windows. Once this hits Stable (should be next week), we will no longer need this workaround.

kayrus:
@joshix not yet.

joshix:
> Locksmith 0.3.4 has support for reboot windows. Once this hits Stable (should be next week), we will no longer need this workaround.

Has this become true?

crawford:
> Has this become true?

Yes.

joshix:
Closing in favor of stock reboot window support in later versions of locksmith (0.3.4+)

PR #643, opened 2015-11-02
Change file permission to '0644'

dghubble:
Projects like https://github.com/coreos/coreos-vagrant YAML.load and YAML.dump the cloud-config which changes octal 0644 to decimal 420, which is inconvenient if you prefer configs consistently use the same radix.


crawford:
Thank you!

PR #647, opened 2015-11-05
os: booting-with-qemu: add instructions for gpg verify

philips:
We do this for PXE, do it for QEMU too: https://coreos.com/os/docs/latest/booting-with-pxe.html


crawford:
LGTM

PR #657, opened 2015-11-06
Correcting the ECS Agent systemd unit file.

JasonSwindle:


bcwaldon:
@JasonSwindle Thanks for the PR! I'm not super familiar with the ECS agent, so please pardon my ignorance. What exactly justifies these changes?

JasonSwindle:
Great question. This change makes sure that the new CloudWatch Metrics[1] get reported correctly. Also, the old version of the unit file was missing some flag / switches the ec-agent docs have.[2]  This will bring the ECS docs and the CoreOS example inline.


[1] http://docs.aws.amazon.com/AmazonECS/latest/developerguide/cloudwatch-metrics.html
[2] http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html

JasonSwindle:
I see an error, let me fix it real fast. :)

JasonSwindle:
Ok, there is the fix. Sorry about that error. :/

JasonSwindle:
Going to re-do this PR, to make it overall cleaner. Sorry about that! :)

JasonSwindle:
https://github.com/coreos/docs/pull/658

PR #658, opened 2015-11-06
Correcting the ECS Agent systemd unit file.

JasonSwindle:
This change makes sure that the new CloudWatch Metrics[1] get reported correctly. Also, the old version of the unit file was missing some flag / switches the ec-agent docs have.[2] This will bring the ECS docs and the CoreOS example inline.

[1] http://docs.aws.amazon.com/AmazonECS/latest/developerguide/cloudwatch-metrics.html
[2] http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html

### DIFF
--- a/os/booting-on-ecs.md
+++ b/os/booting-on-ecs.md
@@ -21,21 +21,37 @@ coreos:
      runtime: true
      content: |
        [Unit]
-       Description=Amazon ECS Agent
-       After=docker.service
-       Requires=docker.service
-       Requires=network-online.target
-       After=network-online.target
-       
+       Description=AWS ECS Agent (https://github.com/aws/amazon-ecs-agent)
+       Documentation=https://docs.aws.amazon.com/AmazonECS/latest/developerguide/
+       Requires=network-online.target docker.socket
+       After=network.target docker.socket
+
        [Service]
        Environment=ECS_CLUSTER=your_cluster_name
-       Environment=ECS_LOGLEVEL=warn
-       Environment=ECS_CHECKPOINT=true
+       Environment=ECS_LOGLEVEL=info
+       Environment=ECS_VERSION=latest
+       Restart=on-failure
+       RestartSec=30
+       RestartPreventExitStatus=5
+       SyslogIdentifier=ecs-agent
+       ExecStartPre=-/bin/mkdir -p /var/log/ecs /var/ecs-data /etc/ecs
+       ExecStartPre=-/usr/bin/touch /etc/ecs/ecs.config
        ExecStartPre=-/usr/bin/docker kill ecs-agent
        ExecStartPre=-/usr/bin/docker rm ecs-agent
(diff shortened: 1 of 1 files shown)


joshix:
LGTM, needs eng review.

os/booting-on-ecs.md
```diff
@@ -21,21 +21,37 @@ coreos:
      runtime: true
      content: |
        [Unit]
-       Description=Amazon ECS Agent
-       After=docker.service
-       Requires=docker.service
-       Requires=network-online.target
-       After=network-online.target
-       
+       Description=AWS ECS Agent (https://github.com/aws/amazon-ecs-agent)
```
crawford: Please remove the URL from the description so that it doesn't clutter up the logs.

os/booting-on-ecs.md
```diff
@@ -21,21 +21,37 @@ coreos:
...
+       Documentation=https://docs.aws.amazon.com/AmazonECS/latest/developerguide/
+       Requires=network-online.target docker.socket
```
crawford: Why does this need to wait on the network? It should restart cleanly.

JasonSwindle: My apologies as I may not be understanding systemd correctly, but without Docker and Networking the ecs-agent is totally worthless. The ecs-agent will not be able to register-container-instance, pull docker images, and etc.   Should I change:

```
After=network.target docker.socket
````

to 

```
After=network-online.target docker.socket
```

Again, my apologies, I am still new to systemd. :)

Comment (crawford) [inline os/booting-on-ecs.md]: Well, my only concern is that "network-online.target" is kind of a hack and is a little racy. The better solution is to create a service which is tolerant of a network failure and can retry. In this case, the service is already tolerant to these failures (if docker fails to pull the image, the service will restart after 30 seconds and the ECS agent itself should be tolerant to these failures as well). I think it is safe to omit the dependency entirely.

The other thing to consider is that the Docker service itself should declare its dependency on the network. Because you are then depending on Docker, the dependency on the network is implicit.

**Comment by JasonSwindle** [comment]
Made the corrections you asked for, thank you on the feedback. :)

File: os/booting-on-ecs.md [hunk 3]
```diff
@@ -21,21 +21,37 @@ coreos:
...
        [Unit]
-       Description=Amazon ECS Agent
-       After=docker.service
-       Requires=docker.service
-       Requires=network-online.target
-       After=network-online.target
-       
+       Description=AWS ECS Agent
+       Documentation=https://docs.aws.amazon.com/AmazonECS/latest/developerguide/
+       Requires=network-online.target docker.socket
```
crawford: You can also remove network-online.target from here as well.

JasonSwindle:
Done. :)

crawford:
Awesome! LGTM

PR #660, opened 2015-11-09
etcd/https-migration: Language rewrite (non-content).

joshix:
Polish language. @kayrus please look over to ensure I haven't rendered anything inaccurate in the process.

(@omkensey and I will be further reworking this to merge some of his content from prosvcs.)

### DIFF
--- a/etcd/etcd-live-http-to-https-migration.md
+++ b/etcd/etcd-live-http-to-https-migration.md
@@ -1,73 +1,73 @@
-# etcd Live HTTP to HTTPS migration
+# Enabling HTTPS in an Existing etcd Cluster
 
-If you already have existing etcd cluster which uses insecure HTTP connections and you would like to switch from HTTP to HTTPS and in addition enable TLS certificate authentication, you have to follow instruction below.
+This guide outlines the process of migrating an existing etcd cluster from HTTP communication to encrypted HTTPS. For added security, it also shows how to require TLS client-server certificates to authenticate connections.
 
 ## Prepare Cluster Components
 
-### Check insecure port availability
+### Check Insecure Port Availability
 
-By default fleet and flannel use two etcd client ports: 2379 (IANA port assignment) and 4001 (legacy port). Last one is used for etcd v0.4 backward compatibility. In this example we will use 4001 TCP port for insecure HTTP connections and our etcd members will listen that port only on local interface for security reason. This will allow us to switch transparently to HTTPS configuration without services downtime and simplify configuration
+By default, etcd communicates with clients over two ports: 2379, the current and official IANA port designation, and 4001, for clients who may implement versions of the protocol older than 0.4. We leverage this quirk of legacy support to migrate a running cluster from insecure plain-HTTP communication (on the old port, 4001) to encrypted HTTPS (on the current port, 2379) without cluster downtime. We restrict the legacy communication on port 4001 to the local interface.
 
-If you've configured flannel, fleetd or other local components to use custom or 2379 port only, you have to configure them to use 4001 TCP port.
+If you've configured flannel, fleet, or other components to use custom ports, or 2379 only, they will be reconfigured to use port 4001.
 
-If your etcd configuration doesn't have configured 4001 client port, you have to enable it before we proceed with this manual. For example you use Cloud-Config to configure your CoreOS cluster. In this case you have to retrieve `ETCD_LISTEN_CLIENT_URLS` value from `/run/systemd/system/etcd2.service.d/20-cloudinit.conf` config file:
+If etcd isn't listening on port 4001, it must also be reconfigured. If you used cloud-config to spin up your machines, you can retrieve the `ETCD_LISTEN_CLIENT_URLS` value from `/run/systemd/system/etcd2.service.d/20-cloudinit.conf` to verify the etcd ports:
 
 ```sh
-grep ETCD_LISTEN_CLIENT_URLS /run/systemd/system/etcd2.service.d/20-cloudinit.conf
+$ grep ETCD_LISTEN_CLIENT_URLS /run/systemd/system/etcd2.service.d/20-cloudinit.conf
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379"
 ```
 
-Then create `/etc/systemd/system/etcd2.service.d/` directory if it doesn't exist and create [drop-in][drop-ins] file: `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. Use value we've retrieved above and append `http://127.0.0.1:4001` URL at the end:
+In this case etcd is listening only on port 2379. We'll add port 4001 with a systemd [drop-in][drop-ins] unit file. Create the file `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. In this file, write an excerpt that appends the new URL on port 4001 to the existing value we retrieved in the step above:
 
 ```
 [Service]
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379,http://127.0.0.1:4001"
 ```
 
-Then run `systemctl daemon-reload` and `systemctl restart etcd2` to restart etcd. Check cluster availability using [`etcdctl`][etcdctl] commands listed below:
+Run `systemctl daemon-reload` followed by `systemctl restart etcd2` to restart etcd. Check cluster status using the [`etcdctl`][etcdctl] commands:
 
 ```sh
-etcdctl member list
-etcdctl cluster-health
+$ etcdctl member list
+$ etcdctl cluster-health
 ```
 
-Repeat these steps on each CoreOS host which has etcd member.
+Repeat these steps on each etcd cluster member.
 
-### Generate TLS keypairs
+### Generate TLS Key Pairs
 
-Follow [generate self signed certificates][self-signed-ca] doc to generate self-signed certificates.
+Follow the guide to [generating self-signed certificates][self-signed-ca] to create the certificate/key pairs needed for each etcd cluster member.
 
-### Copy Keypairs
+### Copy Key Pairs to Nodes
 
-Let's assume we have three CoreOS hosts with three etcd cluster members: server1, server2, server3 and corresponding IP addresses: 172.16.0.101, 172.16.0.102, 172.16.0.103. We will use following keypairs filenames in our example:
+Assume we have three CoreOS machines, running three etcd cluster members: server1, server2, server3; with corresponding IP addresses: 172.16.0.101, 172.16.0.102, and 172.16.0.103. We will use the following key pair file names in our example:
 
 ```
 server1.pem
 server1-key.pem
 ```
 
-Create `/etc/etcd2` directory, copy there corresponding certificate and key and set file permissions:
+Create the `/etc/etcd2` directory, then copy the corresponding certificate and key there. Set permissions to secure the directory and key file:
 
 ```
-chown -R etcd:etcd /etc/etcd2/
-chmod 600 /etc/etcd2/*-key.pem
+$ chown -R etcd:etcd /etc/etcd2
+$ chmod 600 /etc/etcd2/*-key.pem
 ```
 
-Copy `ca.pem` CA certificate into `/etc/etcd2`.
+Copy the `ca.pem` CA certificate file into `/etc/etcd2` as well.
 
-Or copy your `ca.pem` CA certificate into `/etc/ssl/certs` directory and run `update-ca-certificates` script to update your certificates bundle. In this case CA will be available system wide and there will be not necessary to set CA path for each app.
+Alternatively, copy `ca.pem` into `/etc/ssl/certs` instead, and run the `update-ca-certificates` script to update the system certificates bundle. After doing so, the added CA will be available to any program running on the node, and it will not be necessary to set the CA path for each application.
 
-Repeat this step on rest of the member nodes.
+Repeat this step on the rest of the cluster members.
 
-### Using etcd Proxy
+### Using an etcd Proxy
 
-If you typically connect to an etcd cluster at a remote location, it's recommended to use this opportunity to configure an [etcd proxy][etcd proxy] that handles the remote connection logic and TLS termination, and reconfigure your apps to communicate through the proxy on localhost. In this case you have to generate client keypair (i.e. `client1.pem` and `client1-key.pem`) and follow the previous step.
+If you typically connect to a remote etcd cluster, this is a good time to configure an [etcd proxy][etcd proxy] that handles the remote connection and TLS termination, and to reconfigure your apps to communicate through the proxy on localhost. In this case, you must generate a client key pair (e.g. `client1.pem` and `client1-key.pem`) and repeat the *Copy Key Pairs* step above with the client key pair files.
 
-In addition it is necessary to modify your systemd [unit files][systemd-unit-file] or [drop-ins][drop-ins] which use `etcdctl` in `ExecStart*=` or `ExecStop*=` directives and replace `/usr/bin/etcdctl` with the `/usr/bin/etcdctl --no-sync`. This will force `etcdctl` to use proxy for all operations (not only members list).
+It is also necessary to modify your systemd [unit files][systemd-unit-file] or [drop-ins][drop-ins] which use `etcdctl` in `ExecStart*=` or `ExecStop*=` directives to replace the invocation of `/usr/bin/etcdctl` with `/usr/bin/etcdctl --no-sync`. This will force `etcdctl` to use the proxy for all operations.
 
-## Configure etcd Keypair
+## Configure etcd Key Pair
 
-Now we should configure etcd paths to the new certificates. Create `/etc/systemd/system/etcd2.service.d/30-certs.conf` [drop-in][drop-ins] with contents below:
+Now we will configure etcd to use the new certificates. Create a `/etc/systemd/system/etcd2.service.d/30-certs.conf` [drop-in][drop-ins] file with the following contents:
 
 ```
 [Service]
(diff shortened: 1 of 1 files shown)


kayrus:
LGTM

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
-# etcd Live HTTP to HTTPS migration
+# Enabling HTTPS in an Existing etcd Cluster
```
omkensey: General comments: as discussed, I think we should have two versions of this, one tracking 0.x and one tracking 2.x.  Also I just noticed that this doc is in coreos/docs rather than coreos/etc/Documentation -- I think it really belongs in the latter place.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
 ## Prepare Cluster Components
 
-### Check insecure port availability
+### Check Insecure Port Availability
 
-By default fleet and flannel use two etcd client ports: 2379 (IANA port assignment) and 4001 (legacy port). Last one is used for etcd v0.4 backward compatibility. In this example we will use 4001 TCP port for insecure HTTP connections and our etcd members will listen that port only on local interface for security reason. This will allow us to switch transparently to HTTPS configuration without services downtime and simplify configuration
+By default, etcd communicates with clients over two ports: 2379, the current and official IANA port designation, and 4001, for clients who may implement versions of the protocol older than 0.4. We leverage this quirk of legacy support to migrate a running cluster from insecure plain-HTTP communication (on the old port, 4001) to encrypted HTTPS (on the current port, 2379) without cluster downtime. We restrict the legacy communication on port 4001 to the local interface.
```
omkensey: Per previous discussion, if we can avoid using or even mentioning legacy ports in the 2.x branch of the doc, we should.  If it needs to be there for 0.x, so be it, but if not I think we should avoid it there too.  At most, maybe a general mention that "insecure access can be preserved by retaining/adding http:// URLs in the set of listening URLs".

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
-If you've configured flannel, fleetd or other local components to use custom or 2379 port only, you have to configure them to use 4001 TCP port.
+If you've configured flannel, fleet, or other components to use custom ports, or 2379 only, they will be reconfigured to use port 4001.
```
omkensey: I think this is not necessary any more, if it ever was -- I believe I recall somebody on the etcd team saying that fleet and flannel can talk https: to etcd.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
-If your etcd configuration doesn't have configured 4001 client port, you have to enable it before we proceed with this manual. For example you use Cloud-Config to configure your CoreOS cluster. In this case you have to retrieve `ETCD_LISTEN_CLIENT_URLS` value from `/run/systemd/system/etcd2.service.d/20-cloudinit.conf` config file:
+If etcd isn't listening on port 4001, it must also be reconfigured. If you used cloud-config to spin up your machines, you can retrieve the `ETCD_LISTEN_CLIENT_URLS` value from `/run/systemd/system/etcd2.service.d/20-cloudinit.conf` to verify the etcd ports:
```
omkensey: Per above, we may be able to leave out any reference to using 4001 for insecure operation.  From here on I'll just tag "4001" on any lines to refer to this issue.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
 ```sh
-grep ETCD_LISTEN_CLIENT_URLS /run/systemd/system/etcd2.service.d/20-cloudinit.conf
+$ grep ETCD_LISTEN_CLIENT_URLS /run/systemd/system/etcd2.service.d/20-cloudinit.conf
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379"
 ```
 
-Then create `/etc/systemd/system/etcd2.service.d/` directory if it doesn't exist and create [drop-in][drop-ins] file: `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. Use value we've retrieved above and append `http://127.0.0.1:4001` URL at the end:
+In this case etcd is listening only on port 2379. We'll add port 4001 with a systemd [drop-in][drop-ins] unit file. Create the file `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. In this file, write an excerpt that appends the new URL on port 4001 to the existing value we retrieved in the step above:
```
omkensey: 4001

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
 ```
 [Service]
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379,http://127.0.0.1:4001"
 ```
 
-Then run `systemctl daemon-reload` and `systemctl restart etcd2` to restart etcd. Check cluster availability using [`etcdctl`][etcdctl] commands listed below:
+Run `systemctl daemon-reload` followed by `systemctl restart etcd2` to restart etcd. Check cluster status using the [`etcdctl`][etcdctl] commands:
```
omkensey: 4001

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 ```
 
-Repeat these steps on each CoreOS host which has etcd member.
+Repeat these steps on each etcd cluster member.
 
-### Generate TLS keypairs
+### Generate TLS Key Pairs
 
-Follow [generate self signed certificates][self-signed-ca] doc to generate self-signed certificates.
+Follow the guide to [generating self-signed certificates][self-signed-ca] to create the certificate/key pairs needed for each etcd cluster member.
```
omkensey: I understand self-signed is easy for users, but I'd like to either write in here, or link to, a recipe for setting up and running your own CA.  a) that's more secure b) it will track better with the process they follow to get a third-party-CA-signed cert.  That can be put in the linked doc maybe, at the top.  Then this link text would become just "generating certificates".

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
-Let's assume we have three CoreOS hosts with three etcd cluster members: server1, server2, server3 and corresponding IP addresses: 172.16.0.101, 172.16.0.102, 172.16.0.103. We will use following keypairs filenames in our example:
+Assume we have three CoreOS machines, running three etcd cluster members: server1, server2, server3; with corresponding IP addresses: 172.16.0.101, 172.16.0.102, and 172.16.0.103. We will use the following key pair file names in our example:
 
 ```
 server1.pem
 server1-key.pem
 ```
 
-Create `/etc/etcd2` directory, copy there corresponding certificate and key and set file permissions:
+Create the `/etc/etcd2` directory, then copy the corresponding certificate and key there. Set permissions to secure the directory and key file:
```
omkensey: I'd like to have users actually generate the key pairs on the server, rather than shove them around the network, but don't know offhand if this is possible.  If it's not we should investigate adding the necessary tools to the distro.  This may be easy though since apparently OpenSSL and OpenSSH generate mutually-compatible keys.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
-Repeat this step on rest of the member nodes.
+Repeat this step on the rest of the cluster members.
 
-### Using etcd Proxy
+### Using an etcd Proxy
 
-If you typically connect to an etcd cluster at a remote location, it's recommended to use this opportunity to configure an [etcd proxy][etcd proxy] that handles the remote connection logic and TLS termination, and reconfigure your apps to communicate through the proxy on localhost. In this case you have to generate client keypair (i.e. `client1.pem` and `client1-key.pem`) and follow the previous step.
+If you typically connect to a remote etcd cluster, this is a good time to configure an [etcd proxy][etcd proxy] that handles the remote connection and TLS termination, and to reconfigure your apps to communicate through the proxy on localhost. In this case, you must generate a client key pair (e.g. `client1.pem` and `client1-key.pem`) and repeat the *Copy Key Pairs* step above with the client key pair files.
```
omkensey: I'm not clear on what this means -- the proxy is going to accept insecure client connections and connect securely to the cluster?

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
 
-In addition it is necessary to modify your systemd [unit files][systemd-unit-file] or [drop-ins][drop-ins] which use `etcdctl` in `ExecStart*=` or `ExecStop*=` directives and replace `/usr/bin/etcdctl` with the `/usr/bin/etcdctl --no-sync`. This will force `etcdctl` to use proxy for all operations (not only members list).
+It is also necessary to modify your systemd [unit files][systemd-unit-file] or [drop-ins][drop-ins] which use `etcdctl` in `ExecStart*=` or `ExecStop*=` directives to replace the invocation of `/usr/bin/etcdctl` with `/usr/bin/etcdctl --no-sync`. This will force `etcdctl` to use the proxy for all operations.
 
-## Configure etcd Keypair
+## Configure etcd Key Pair
 
-Now we should configure etcd paths to the new certificates. Create `/etc/systemd/system/etcd2.service.d/30-certs.conf` [drop-in][drop-ins] with contents below:
+Now we will configure etcd to use the new certificates. Create a `/etc/systemd/system/etcd2.service.d/30-certs.conf` [drop-in][drop-ins] file with the following contents:
```
omkensey: Is this necessary if we put the cert(s) in the system bundle or only if we copied them to /etc/etcd2?

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -81,18 +81,18 @@ Environment="ETCD_PEER_TRUSTED_CA_FILE=/etc/etcd2/ca.pem"
...
 ```
 
-If everything is ok, repeat this step on rest of the member nodes.
+Repeat this step on the rest of the cluster members.
 
-### Configure etcd Proxy
+### Configure etcd Proxy Key Pair
 
-In the case of using an etcd proxy you have to use `/etc/systemd/system/etcd2.service.d/30-certs.conf` [drop-in][drop-ins] with the contents below:
+If proxying etcd connections as discussed above, create a systemd [drop-in][drop-ins] unit file named `/etc/systemd/system/etcd2.service.d/30-certs.conf` with the following contents:
```
omkensey: Ditto the previous comment.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -102,39 +102,29 @@ Environment="ETCD_TRUSTED_CA_FILE=/etc/etcd2/ca.pem"
...
 
 ```sh
-curl http://127.0.0.1:4001/v2/stats/self
+$ curl http://127.0.0.1:4001/v2/stats/self
 ```
 
 ## Change etcd Peer URLs
 
-The command below should generate commands which you have to run manually on one of the member nodes:
+Use `etcdctl` piped through an `awk` filter to print the commands needed to reconfigure the cluster peer URLs. After reviewing the command lines, we will invoke them on one etcd cluster member.
```
omkensey: I really, really, really hate just telling people to run awk/sed/bash transform scripts for any but the most trivial operations.  I'd rather just tell them the change to make, and let them decide how to make it.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -102,39 +102,29 @@ Environment="ETCD_TRUSTED_CA_FILE=/etc/etcd2/ca.pem"
...
-```
-
-Output should look like:
-
-```sh
-curl -XPUT -H "Content-Type: application/json" http://localhost:2379/v2/members/2428639343c1baab -d '{"peerURLs":["https://172.16.0.102:2380"]}'
-curl -XPUT -H "Content-Type: application/json" http://localhost:2379/v2/members/50da8780fd6c8919 -d '{"peerURLs":["https://172.16.0.103:2380"]}'
-curl -XPUT -H "Content-Type: application/json" http://localhost:2379/v2/members/81901418ed658b78 -d '{"peerURLs":["https://172.16.0.101:2380"]}'
-```
+### Current etcd
```
omkensey: Per previous discussion we'd be splitting these.  I think that will especially make this section clearer as it interleaves operations A and B for etcd-current, A and B for etcd-old, and then finishes up with an operation to be performed in either case, which is confusing.  Without the interleaving it will just be a linear set of steps.

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -142,24 +132,41 @@ etcdctl member update 50da8780fd6c8919 https://172.16.0.103:2380
 etcdctl member update 81901418ed658b78 https://172.16.0.101:2380
 ```
 
-You have to run generated commands **except last one** on one of the member nodes and check etcd cluster health:
+On any one etcd cluster member, run each of the printed commands -- *except the last one*. We will change the peer URL for the last etcd cluster member only after completing any proxy configuration below.
```
omkensey: Why do we need to wait on the etcd proxy?  *Do* we need to?  We should explain the reason(s) better, or if (as I suspect), we don't actually need to delay, then just plow ahead.

etcd/etcd-live-http-to-https-migration.md
kayrus: This doc is in os because we use CoreOS related stuff here. Systemd drop-ins, `etcd2`, `update-ca-certificates`, etc.

etcd/etcd-live-http-to-https-migration.md
kayrus: We have to use legacy port here because we keep it insecure and other CoreOS apps use it by default and we don't need to reconfigure them.

etcd/etcd-live-http-to-https-migration.md
kayrus: They can. But you have to reconfigure them and this doc will be much longer. In addition we have to restart fleet with a hook `kill -9 $(pidof fleetd); systemctl restart fleet` to keep existing scheduled units. And I would like to avoid that.

etcd/etcd-live-http-to-https-migration.md
kayrus: What is wrong with existing link? It points here: https://github.com/coreos/docs/blob/master/os/generate-self-signed-certificates.md and it explains what is own CA and so on. Probably we have to improve that doc?

etcd/etcd-live-http-to-https-migration.md
kayrus: Well, if user will generate certificates directly on the server it means that server should contain CA private key. It is potentially insecure.

Previously we have discussed which tool should we use, openssl or cfssl. As I understood we have to use cfssl. If we've changed our minds, let's rewrite [this](https://github.com/coreos/docs/blob/master/os/generate-self-signed-certificates.md) doc.

etcd/etcd-live-http-to-https-migration.md
kayrus: Exactly. It will accept insecure connections on localhost. And again, I've made that to avoid potential services downtime.

etcd/etcd-live-http-to-https-migration.md
kayrus: CA bundle contains only CA certs

etcd/etcd-live-http-to-https-migration.md
kayrus: I agree with you, but even I can make a mistake during following this doc. Copy-paste wrong server for example. As a workaround we can add additional explanation for these strings.

etcd/etcd-live-http-to-https-migration.md
kayrus: If we switch URLs of all etcd members, and don't reconfigure them in `proxy-refresh-interval` (30 sec) time period, etcd proxy will update these urls and will try to connect to new URLs which are not available yet. That will break cluster. If we keep one etcd member to listen insecure port - etcd proxy will still work.

We cannot accomplish this procedure in 30 secs. Or even if we increase this interval - something could get wrong and we'll have troubles.

etcd/etcd-live-http-to-https-migration.md
kayrus: etcd 0.4 doesn't support HTTPS. There is nothing to split.

kayrus:
@omkensey @joshix any additional notes on this PR?

kayrus:
@joshix for flannel compatibility we have to replace "/etc/etcd2/" directory with the "/etc/ssl/etcd/".

etcd/etcd-live-http-to-https-migration.md
```diff
@@ -1,73 +1,73 @@
...
+$ grep ETCD_LISTEN_CLIENT_URLS /run/systemd/system/etcd2.service.d/20-cloudinit.conf
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379"
 ```
 
-Then create `/etc/systemd/system/etcd2.service.d/` directory if it doesn't exist and create [drop-in][drop-ins] file: `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. Use value we've retrieved above and append `http://127.0.0.1:4001` URL at the end:
+In this case etcd is listening only on port 2379. We'll add port 4001 with a systemd [drop-in][drop-ins] unit file. Create the file `/etc/systemd/system/etcd2.service.d/25-insecure_localhost.conf`. In this file, write an excerpt that appends the new URL on port 4001 to the existing value we retrieved in the step above:
 
 ```
 [Service]
 Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379,http://127.0.0.1:4001"
```
kayrus: @joshix there is a typo. please replace http://0.0.0.0:2379 with the "http://127.0.0.1:2379"

joshix:
@kayrus Updated path to `/etc/ssl/etcd/`.

PR #665, opened 2015-11-11
os/selinux: s/realpath/readlink -f/

joshix:
- Change progression: readlink (corner failures) -> realpath (obsolescent) ->
readlink -f (desired canonicalization of symlink absolute path).
- Link the file-issue recomendation to the bugs/issues URL.
- Title Case Headings.
- Tune language/style.
- Prefix blocked cmd lines with prompt (`$`) to distinguish from cmd output.

### DIFF
--- a/os/selinux.md
+++ b/os/selinux.md
@@ -1,12 +1,16 @@
 # Using SELinux on CoreOS
 
-SELinux is a fine-grained access control mechanism that is integrated into
-CoreOS and rkt. Each container is run in its own independent SELinux
-context, increasing isolation between containers even in the event of a
-container being compromised. CoreOS implements SELinux, but currently does
-not enforce SELinux protections by default.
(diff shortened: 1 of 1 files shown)


joshix:
@kayrus Addendum PR to your merge. I switched to `readlink -f`. Objections/PTAL?

os/selinux.md
```diff
@@ -16,22 +20,25 @@ enable SELinux logging by running the following commands as root:
 * `semodule -DB`
 * `systemctl restart audit-rules`
```
kayrus: Should we add `$` here too?

kayrus:
Added one note

os/selinux.md
joshix: Yep. Good catch. Will fix squash merge shortly.

joshix:
Actually will merge this, *then* fix the above comment and a few other bits.

PR #666, opened 2015-11-11
Fix SELinux debug instructions

mjg59:
/usr is read-only, so semodule can't rebuild the policy database. Update
the instructions so the policy store is copied to a writable filesystem


joshix:
Aha. Nicely fielded. LGTM.

PR #661, opened 2015-11-09
Fixed CLI for SElinux doc

kayrus:
`readlink` didn't provide proper result when you are inside depth > 2 path.

PR #668, opened 2015-11-12
os/selinux: Make more consistent with other docs; explain cmds.

joshix:
Eliminate the bullet points per step that were unlike other docs.
Minimally explicate the commands we're running.
Prefix all cmd lines with the prompt `$` to distinguish I from O.

(Finishes up what started in #661, #665, and #666.)

### DIFF
--- a/os/selinux.md
+++ b/os/selinux.md
@@ -1,4 +1,4 @@
-# Using SELinux on CoreOS
+# SELinux on CoreOS
 
 SELinux is a fine-grained access control mechanism integrated into
 CoreOS and rkt. Each container runs in its own independent SELinux context,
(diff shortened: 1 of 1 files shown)


joshix:
@mjg59 PTAL to ensure I didn't foul anything while making this doc a little more complete and consistent.
@kayrus Here is the last of the selinux doc series between the three of us.

kayrus:
These `$` don't allow user to copy-paste all commands at once. Someone says that it is a point, but I'd say nope.

joshix:
Couldn't have copied those lines at once before these changes either: They were each prefixed with a bullet.

kayrus:
@joshix nope, when you copy-paste into terminal bullets are ignored

joshix:
Interesting. But which terminal, which shell?
In general, we have this idea that we want to show expected output for many commands. That output also prevents mass cut and paste, and the prompt distinguishes it from the command line above and/or below it.

@robzumski unless you strenuously object , I'm inclined to merge this and continue applying the shell prompt throughout the docs.

PR #669, opened 2015-11-13
qe: Add ImagePullSecret to machine config doc

robszumski:
cc: @joshix

### DIFF
--- a/quay-enterprise/configure-machines.md
+++ b/quay-enterprise/configure-machines.md
@@ -25,7 +25,7 @@ eyAiaHR0cHM6Ly9pbmRleC5kb2NrZXIuaW8vdjEvIjogeyAiYXV0aCI6ICJabUZyWlhCaGMzTjNiM0pr
 apiVersion: v1
 kind: Secret
 metadata:
-  name: FoobarQuayCreds
+  name: foobarquaycreds
 data:
   .dockercfg: eyAiaHR0cHM6Ly9pbmRleC5kb2NrZXIuaW8vdjEvIjogeyAiYXV0aCI6ICJabUZyWlhCaGMzTjNiM0prTVRJSyIsICJlbWFpbCI6ICJqZG9lQGV4YW1wbGUuY29tIiB9IH0K
 type: kubernetes.io/dockercfg
@@ -35,7 +35,7 @@ To use this secret, first submit it into the cluster:
 
 ```sh
 $ kubectl create -f /tmp/foobarquaycreds.yaml
-secrets/FoobarQuayCreds
+secrets/foobarquaycreds
 ```
 
 And then reference it in a Pod or Replication Controller:
@@ -44,13 +44,13 @@ And then reference it in a Pod or Replication Controller:
 apiVersion: v1
 kind: Pod
 metadata:
-  name: Foobar
+  name: foobar
 spec:
   containers:
     - name: foo
       image: quay.io/coreos/etcd:v2.2.1
   imagePullSecrets:
-    - name: FoobarQuayCreds
+    - name: foobarquaycreds
 ```
 
 ### Cloud-Config
--- a/quay-enterprise/configure-machines.md
+++ b/quay-enterprise/configure-machines.md
@@ -38,7 +38,7 @@ $ kubectl create -f /tmp/foobarquaycreds.yaml
 secrets/foobarquaycreds
 ```
 
-And then reference it in a Pod or Replication Controller:
+And then reference it in a Pod:
 
 ```
 apiVersion: v1
@@ -51,6 +51,31 @@ spec:
       image: quay.io/coreos/etcd:v2.2.1
   imagePullSecrets:
     - name: foobarquaycreds
+```
+
+Or a Replication Controller:
+
+```
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: foobar
+spec:
+  replicas: 2
+  selector:
+    app: fooapp
+  template:
+    metadata:
+      labels:
+        app: fooapp
+    spec:
+      containers:
+        - name: foo
+          image: quay.io/coreos/etcd:v2.2.1
+          ports:
+            - containerPort: 80
+      imagePullSecrets:
+        - name: foobarquaycreds
 ```
 
 ### Cloud-Config
--- a/quay-enterprise/configure-machines.md
+++ b/quay-enterprise/configure-machines.md
@@ -12,7 +12,7 @@ Writing the `.dockercfg` can be specified in [cloud-config](https://coreos.com/o
 
 ### Kubernetes Pull Secret
 
-If you are using Quay Enterprise in conjunction with a Kubernetes or Tectonic cluster, it's easiest to use the built in secret distribution method. This method allows for you to use different sets of robot accounts on a per-app basis, and also allows for them to be updated or rotated at any time across all machines in the cluster.
+If you are using Quay Enterprise in conjunction with a Kubernetes or Tectonic cluster, it's easiest to use the built-in secret distribution method. This method allows for you to use different sets of robot accounts on a per-app basis, and also allows for them to be updated or rotated at any time across all machines in the cluster.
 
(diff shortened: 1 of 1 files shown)


kbrwn:
I think that using capitals for the name (ie FoobarQuayCreds) will cause a regex error. Should change it to fooquaycreds.

robszumski:
@kbrwn Good call, updated.

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,58 @@
...
+Each CoreOS machine needs to be configured with the username and password for a robot account in order to deploy your containers. Docker looks for configured credentials in a `.dockercfg` file located within the user's home directory. You can download this file directly from the Quay Enterprise interface. Let's assume you've created a robot account called `myapp+deployment`.
 
 Writing the `.dockercfg` can be specified in [cloud-config](https://coreos.com/os/docs/latest/cloud-config.html) with the write_files parameter, or created manually on each machine.
 
+### Kubernetes Pull Secret
+
+If you are using Quay Enterprise in conjunction with a Kubernetes or Tectonic cluster, it's easiest to use the built in secret distribution method. This method allows for you to use different sets of robot accounts on a per-app basis, and also allows for them to be updated or rotated at any time across all machines in the cluster.
```
joshix: ...use the built-in secret distribution method ...

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,58 @@
...
+```
+
+To use this secret, first submit it into the cluster:
+
+```sh
+$ kubectl create -f /tmp/foobarquaycreds.yaml
+secrets/foobarquaycreds
+```
+
+And then reference it in a Pod or Replication Controller:
```
joshix: ...Pod or Replication Controller YAML definition:

(is that clearer?)

joshix:
Two quibbles, otherwise LGTM

kbrwn:
An alternative is to edit the default service account to use a imagePullSecret:
$ kubectl get serviceaccounts default -o yaml > ./sa.yaml
$ vi sa.yaml
delete line with key "resourceVersion"
add imagePullSecret
$ kubectl replace serviceaccount default -f ./sa.yaml

quay-enterprise/configure-machines.md
jonboulle: I like this guy

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,58 @@
...
 Writing the `.dockercfg` can be specified in [cloud-config](https://coreos.com/os/docs/latest/cloud-config.html) with the write_files parameter, or created manually on each machine.
 
+### Kubernetes Pull Secret
+
+If you are using Quay Enterprise in conjunction with a Kubernetes or Tectonic cluster, it's easiest to use the built in secret distribution method. This method allows for you to use different sets of robot accounts on a per-app basis, and also allows for them to be updated or rotated at any time across all machines in the cluster.
+
+An "Image Pull Secret" is a special secret that Kubernetes will use when pulling down the containers in a pod. It is a base64-encoded Docker config file. Here's an example:
+
+```sh
+$ cat ~/.dockercfg | base64
```
jonboulle: base64 ~/.dockercfg, save a fork, save a tree

robszumski: does this work on both mac/linux?

jonboulle: ugh, I suppose not -
https://developer.apple.com/library/prerelease/mac/documentation/Darwin/Reference/ManPages/man1/base64.1.html

base64 < ~/.dockercfg should though.

robszumski:
@kbrwn Check out the latest commit

kbrwn:
lgtm

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,74 @@
-# Configure Machines for Enterprise Registry
+# Configure Machines for Quay Enterprise
 
-The Enterprise Registry allows you to create teams and user accounts that match your existing business unit organization. A special type of user, a robot account, is designed to be used programatically by deployment systems and other pieces of software. Robot accounts are commonly configured with read-only access to an organizations repositories.
+Quay Enterprise allows you to create teams and user accounts that match your existing business unit organization. A special type of user, a robot account, is designed to be used programatically by deployment systems and other pieces of software. Robot accounts are commonly configured with read-only access to an organizations repositories.
 
 This guide we will assume you have the DNS record `registry.example.com` configured to point to your Enterprise Registry.
```
joshix: This guide assumes the DNS A record `registry.example.com` points to your Quay Enterprise Registry.

("Quay" may have been intentionally committed in your version and if so, OK, omit it here.

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,74 @@
-# Configure Machines for Enterprise Registry
+# Configure Machines for Quay Enterprise
 
-The Enterprise Registry allows you to create teams and user accounts that match your existing business unit organization. A special type of user, a robot account, is designed to be used programatically by deployment systems and other pieces of software. Robot accounts are commonly configured with read-only access to an organizations repositories.
+Quay Enterprise allows you to create teams and user accounts that match your existing business unit organization. A special type of user, a robot account, is designed to be used programatically by deployment systems and other pieces of software. Robot accounts are commonly configured with read-only access to an organizations repositories.
```
joshix: Soft suggest:
Quay Enterprise allows you to create user accounts and *teams*, or groups, of those users that mirror your existing org chart.

...
s/organizations/organization's/
OR
"Robot accounts are usually configured with read-only access to a repository."

quay-enterprise/configure-machines.md
```diff
@@ -1,15 +1,74 @@
...
 
 This guide we will assume you have the DNS record `registry.example.com` configured to point to your Enterprise Registry.
 
 ## Credentials
 
-Each CoreOS machine needs to be configured with the username and password for a robot account in order to deploy your containers. Docker looks for configured credentials in a `.dockercfg` file located within the user's home directory. You can download this file directly from the Enterprise Registry interface. Let's assume you've created a robot account called `myapp+deployment`.
+Each CoreOS machine needs to be configured with the username and password for a robot account in order to deploy your containers. Docker looks for configured credentials in a `.dockercfg` file located within the user's home directory. You can download this file directly from the Quay Enterprise interface. Let's assume you've created a robot account called `myapp+deployment`.
```
joshix: IMO, files are best described as "beneath" rather than "within" their parent directory because it avoids a cyclic graph when talking about symbolic links, mounts, and binds. However I'm just kind of marginalia'ing this here; I don't expect an edit but rather a later style guide matter.

quay-enterprise/configure-machines.md
joshix: My /usr/bin/base64 on 10.11.1 seems to read stdin in the absence of args. FWIW.

```
j-mbp:~ j$ echo farnk > .dockercfg
j-mbp:~ j$ which base64
/usr/bin/base64
j-mbp:~ j$ /usr/bin/base64 ~/.dockercfg
ZmFybmsK
```

`base64(1)` says it does so, as well. What am I missing in this discussion?

joshix:
Those few comments, otherwise a second LGTM.

I think later also we should figure out how to kill `foo` `bar` and `baz` forever and everywhere. I think *foobarquaycreds* is rife with negative implications.

robszumski:
@joshix I hate doing that too. Should I say App? Web?

joshix:
@robszumski IIRC in previous comments that foobarquaycreds string needed to be lowercased, because of a dependency in a grep?

joshix:
`websvc` ; `websvcquaycreds`; &c.

`MyUnicornIP` ; `MyUnicornIPquaycreds`; ...

robszumski:
@joshix No, its a requirement of Kubernetes. Very annoying.

quay-enterprise/configure-machines.md
robszumski: I left this the same as the orig FYI

PR #673, opened 2015-11-18
Added static route example

kayrus:
subj


joshix:
LGTM. Merging for a quick lang polish subsequently

PR #677, opened 2015-11-19
os/network-config-networkd: Polish language from #673.

joshix:


os/network-config-with-networkd.md
```diff
@@ -31,7 +31,7 @@ Setting up static networking in your cloud-config can be done by writing out the
 
 ```yaml
 #cloud-config
-
```
joshix: Note to self: my editor is doing this automagically

PR #678, opened 2015-11-20
ignition: update etcd example to use %m

crawford:


mischief:
lgtm, now only if we had a test.. :p

PR #679, opened 2015-11-21
Correct booting-with-qemu.md bzip2 command

nelsonjchen:
If you're going to decompress an image, you need to specify the bz2 file it is from as the filename argument, not the image name.


joshix:
Merged. Thanks for the catch!

PR #684, opened 2015-11-30
os: update note on auto rollbacks on ec2

crawford:


robszumski:
LGTM. Maaaaybe consider putting in the CoreOS version of the change.

PR #687, opened 2015-12-03
os/nginx cloud-config: Polish language in nginx cloud-config docs

joshix:


os/booting-with-libvirt.md
```diff
@@ -71,9 +71,7 @@ mkdir -p /var/lib/libvirt/images/coreos/coreos1/openstack/latest
 touch /var/lib/libvirt/images/coreos/coreos1/openstack/latest/user_data
 ```
 
-The `user_data` file may contain a script for a [cloud config][cloud-config]
-file. We recommend using ssh keys to log into the VM so at a minimum the
-contents of `user_data` should look something like this:
+The `user_data` file declares machine configuration in the [cloud config]({{site.baseurl}}/docs/cluster-management/setup/cloudinit-cloud-config) format. We recommend using ssh keys to log into the VM, and since those keys are stored in `user_data,` at minimum that file should contain something like this:
```
joshix: @robszumski This link to the cloud-config doc is redirected by httpd to land at coreos.com/docs/os/latest/cloud-config.html
Source doc for that actually lives in the [coreos-cloudinit repo at `Documentation/cloud-config.md`](https://github.com/coreos/coreos-cloudinit/blob/master/Documentation/cloud-config.md).

How should it be linked to?

os/nginx-host-cloud-config.md
```diff
@@ -0,0 +1,23 @@
...
+  root /path/to/cloud/config/files;
+  sub_filter $public_ipv4 '$remote_addr';
+  sub_filter $private_ipv4 '$http_x_forwarded_for';
+# sub_filter $private_ipv4 '$http_x_real_ip';
+  sub_filter_once off;
+  sub_filter_types '*';
+}
+```
+
+This example configuration is valid for all `/user_data*` URIs (e.g., `/user_data_host1`, `/user_data_host2`). With a remote nginx accessed via a transparent proxy, `$private_ipv4` substitution will work only if the proxy adds appropriate `HTTP_X_FORWARDED_FOR` or `HTTP_X_REAL_IP` HTTP header to requests.
```
joshix: @kayrus Could you verify the accuracy of these two sentences? I'm not 1000% sure I understood your original perfectly.

os/booting-with-libvirt.md
robszumski: I think we should probably go with a direct link to `https://coreos.com/os/docs/latest/cloud-config.html` since our rewriter isn't smart enough to know that these two repos get merged together.

If we just did `cloud-config.html` as a relative link, it wouldn't work on Github...

PR #688, opened 2015-12-04
ignition/examples: Tweaks to language and format.

joshix:
Unwrap lines. Inline links. ATX headers. Typos fixed. Polish a little language. Ref #633

### DIFF
--- a/ignition/examples.md
+++ b/ignition/examples.md
@@ -1,14 +1,10 @@
-# Example Configs #
+# Example Configs
 
-Each of these examples is written in version 1 of the config. Always double
-check to make sure that your config matches the version that Ignition is
-expecting.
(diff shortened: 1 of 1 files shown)


crawford:
-1

joshix:
@crawford On the unwrap?

crawford:
Yeah. Everything else looks good.

joshix:
Gotcha. I think the churn wrapping causes in prose is just too troublesome. So we're going to have long markdown lines if it's my call. 

I do however understand where you're coming from. In the spirit of gofmt, I think we can all benefit from picking a format even tho that format may not entirely please any one individual. 

I humbly submit that its pretty easy to read these in a terminal with fmt.

crawford:
Died a little inside. Thanks for not going full-retard and doing one sentence per line.

joshix:
I couldn't find my special hat for that.

PR #690, opened 2015-12-05
Update mount location

Malet:
I found that this is no longer being mounted at `/media/root`, I was able to find the files mounted here instead - maybe it needs a symlink?


crawford:
The existing comment is correct. From the point-of-view of toolbox, the host's filesystem is mounted to `/media/root`. From the point-of-view of the host, toolbox's filesystem is located at `/var/lib/toolbox/core-fedora-latest`.

PR #693, opened 2015-12-11
os: digitalocean: remove link to size.png

philips:
This link was broken and probably unessecary.

#hackday 

 cc @sym3tri


robszumski:
LGTM

PR #674, opened 2015-11-18
Added doc on how to configure CoreOS components to use HTTPS connecti…

kayrus:
…on with etcd cluster
/cc @joshix @omkensey

PR #701, opened 2015-12-17
Revert "Added doc on how to configure CoreOS components to use HTTPS connecti…"

joshix:
Reverts coreos/docs#674

WebUI button oops pulled trigged too soon. We need a different file name.


joshix:
I have forgotten how to use git in the space of 12 hours.

PR #706, opened 2015-12-23
suggest ext4 instead of btrfs for docker partition

arthur-c:
After few weeks with btfrs partition for /var/lib/docker I ran into the "No Space Left on Device" btrfs problem mentioned here : https://coreos.com/os/docs/latest/btrfs-troubleshooting.html

This page says "Btrfs is still supported and works with the latest CoreOS releases and Docker, but we recommend using ext4." 

mounting-storage.html shows an example with a btrfs partition for /var/lib/docker, increasing the chance to provoke the "No Space Left on Device" btrfs problem. It might prevent the problem to not suggest btrfs in this case.

### DIFF
--- a/os/mounting-storage.md
+++ b/os/mounting-storage.md
@@ -22,7 +22,7 @@ coreos:
 
 Docker containers can be very large and debugging a build process makes it easy to accumulate hundreds of containers. It's advantageous to use attached storage to expand your capacity for container images. Be aware that some cloud providers treat certain disks as ephemeral and you will lose all Docker images contained on that disk.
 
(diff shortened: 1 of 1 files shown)


joshix:
LGTM and +1 the change. ext4's `-F` is the closest equivalent to btrfs's `-f` (although their respective manuals make ext4's `force` seem more permissive).

@robszumski @crawford: Thoughts or objections?

robszumski:
None. Will defer to @crawford for ongoing docker compatibility questions

crawford:
Looks fine.

PR #709, opened 2015-12-24
Change 'Azure' → 'Microsoft Azure'

ahmetalpbalkan:
Also this needs to be updated at https://coreos.com/os/docs/latest/

image

PR #711, opened 2015-12-27
os/booting-on-azure: Style guide fixes: Unwrap. Ref links.

joshix:


os/booting-on-azure.md
```diff
@@ -1,14 +1,10 @@
 # Running CoreOS on Microsoft Azure
 
-## Choosing a Channel
+## Choosing a channel
```
jonboulle: ??

jonboulle: whoah, I missed that we moved to sentence case. yay!

joshix: Sentence case fits the way these documents tend to use a large number of structural headings.It looks less quaint, and we'll save review time by not trying to figure out the many variations of Title Case (which we retain for titles).

(Watchers, see the [style guide, *Heading style*](https://github.com/coreos/docs/blob/master/STYLE.md#heading-style).)

joshix:
atop and to be synced with #709

PR #716, opened 2015-12-30
os/sdk/modifying: Add and clarify instructions for arm64

joshix:
Continuing on and superseding #595, add instructions for arm64-usr target,
with headings to clearly indicate the context switches for an arm64 build.
Attempt to dispose of the aarch64/arm64 synonyms with an explanation.

### DIFF
--- a/os/sdk-modifying-coreos.md
+++ b/os/sdk-modifying-coreos.md
@@ -1,33 +1,22 @@
 # CoreOS Developer SDK Guide
 
-These are the instructions for building CoreOS itself. By the end of
-the guide you will build a developer image that you can run under
-KVM and have tools for making changes to the code.
+These are the instructions for building CoreOS itself. By the end of the guide you will build a developer image that you can run under KVM and have tools for making changes to the code.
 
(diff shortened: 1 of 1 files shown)


joshix:
cc @marineam @robszumski

Especially, marineam: Are these instructions still current? I haven't attempted an arm64 build.

os/sdk-modifying-coreos.md
```diff
@@ -84,37 +106,67 @@ tooling.
...
+./setup_board --default --board=arm64-usr
+```
+
+#### Compile and link system binaries
+
 Build all of the target binary packages:
 
 ```sh
-./build_packages
+./build_packages [ --board=arm64-usr ]
```
marineam: I'd say leave off the --board arg to keep the examples copy/pasteable because people will copy/paste anyway. You can add a note to the text above that `--board=` may be needed.

joshix: I can see that happening. And the explanatory note about most of the scripts here taking the `--board` option already exists in the graf above. I'll make the suggested change.

PR #721, opened 2016-01-05
ignition: fix raid example

crawford:
/cc @vcaputo


joshix:
LGTM (wipeTable verified in ignition; -L/--label exchange in mkfs.ext4)

Issue #722, opened 2016-01-07
https://coreos.com/docs/launching-containers/launching/getting-started-with-systemd/ hello.service's pull


kaihendry:
Upon https://coreos.com/docs/launching-containers/launching/getting-started-with-systemd/ you have a `hello.service` that has something like `ExecStartPre=/usr/bin/docker pull abiosoft/caddy:php` in my example. However when this fails everything seems to fail.

```
 Jan 07 07:57:38 as.dl.webconverger.com docker[727]: Error while pulling image: Get https://index.docker.io/v1/repositories/abiosoft/caddy/images: dial tcp: lookup index.docker.io: Temporary failure in name resolution
```

Full log: http://s.natalian.org/2016-01-07/caddy-journal.txt

So my workaround has been to comment out that line or fix it to allow it to fail?

```
ExecStartPre=-/usr/bin/docker pull busybox
```

I am testing upon CoreOS 899.1.0 (alpha).

kayrus:
example provided on the https://coreos.com/docs/launching-containers/launching/getting-started-with-systemd/ page has an unobvious exception (I mean systemct enable *unit*) - cloud-providers. we can use "systemctl enable *unit*" only when we are absolutely sure what we are doing. for the common usage we have to use Cloud-Config: https://coreos.com/os/docs/latest/cloud-config.html#units. because cloud-config takes care of network configuration for cloud providers. and when you use systemd's autostart feature ("systemct enable") - that could case an issue like your one. when you use cloud-config to run your units - cloud-config will run them after the network configuration and you'll avoid such problems.

I've added a note in this doc: https://github.com/coreos/docs/pull/723

crawford:
There is nothing inherently wrong with enabling a service. In fact, that will be the preferred approach in a post-cloudinit world. The underlying issue is 1) the service doesn't properly depend on `network.target` and 2) the service doesn't retry if there is a network failure (maybe the network isn't moving packets yet or maybe someone tripped in the datacenter). Now, coming back to cloud-init: I suggest people avoid using `enable` and ensuring `runtime` is set in an effort to make _every_ boot similar. Otherwise, your first boot is going to be very different from subsequent boots (because services are now enabled). This mistake in design is one of the big reasons we are building [Ignition](https://coreos.com/ignition/docs/latest/what-is-ignition.html).

crawford:
Closing due to inactivity.

PR #723, opened 2016-01-07
Added a note about "systemct enable" and cloud providers

kayrus:
Resolves https://github.com/coreos/docs/issues/722

/cc @joshix @robszumski

### DIFF
--- a/os/getting-started-with-systemd.md
+++ b/os/getting-started-with-systemd.md
@@ -8,7 +8,7 @@ systemd consists of two main concepts: a unit and a target. A unit is a configur
 
 systemd is the first process started on CoreOS and it reads different targets and starts the processes specified which allows the operating system to start. The target that you'll interact with is the `multi-user.target` which holds all of the general use unit files for our containers.
 
-Each target is actually a collection of symlinks to our unit files. This is specified in the unit file by `WantedBy=multi-user.target`. Running `systemctl enable foo.service` creates symlinks to the unit inside `multi-user.target.wants`.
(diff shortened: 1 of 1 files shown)


os/getting-started-with-systemd.md
```diff
@@ -8,7 +8,7 @@ systemd consists of two main concepts: a unit and a target. A unit is a configur
 
 systemd is the first process started on CoreOS and it reads different targets and starts the processes specified which allows the operating system to start. The target that you'll interact with is the `multi-user.target` which holds all of the general use unit files for our containers.
 
-Each target is actually a collection of symlinks to our unit files. This is specified in the unit file by `WantedBy=multi-user.target`. Running `systemctl enable foo.service` creates symlinks to the unit inside `multi-user.target.wants`.
+Each target is actually a collection of symlinks to our unit files. This is specified in the unit file by `WantedBy=multi-user.target`. Running `systemctl enable foo.service` creates symlinks to the unit inside `multi-user.target.wants`. Please note that `systemctl enable foo.service` may affect the boot process by race condition exception in network configuration when [Cloud-Config] configures network according to cloud provider's `meta_data` and `user_data`. For the common usage you have to put your unit files inside Cloud-Config's [units][cloud-config-units] section.
```
robszumski: This sentence is really hard to parse/understand, but I don't really have ideas for how to rewrite. @joshix?

kayrus:
These `systemct enable %service%` issues are quite common. This week I already met two users with this problem.

 We have to aware them.

kayrus:
Probably we have to add this info at Cloud-Config doc here: https://github.com/coreos/coreos-cloudinit/blob/master/Documentation/cloud-config.md#units

os/getting-started-with-systemd.md
crawford: I don't like this advice. I think we should tell our users that if they want to use cloud-init, all services should be `runtime: true` and `enabled: false`. This way, every boot is the same.

There is nothing wrong with systemd enable, it is entirely our fault. When people start moving to Ignition, they will use systemd enable to run there services and it will work perfectly. I don't want to mislead people here.

joshix:
closed inactive and addressed otherwise in both cloud-config.md and with ignition switch

PR #724, opened 2016-01-07
Update booting-digitalocean with warning about insecure network

### LINKED ISSUE #718: Insecure private networking with DigitalOcean
I am looking at the tutorial for setting up CoreOS on digitalOcean, and the default is to listen on the private_ip. The problem is that private networking on digitalocean is shared.

Wouldn't that mean that everybody inside the private networking on digitalocean would be able to join my coreos cluster?

kevinsimper:
Closes #718

### DIFF
--- a/os/booting-on-digitalocean.md
+++ b/os/booting-on-digitalocean.md
@@ -7,6 +7,8 @@ schedules per channel. You can [disable this feature][reboot-docs], although we
 don't recommend it. Read the [release notes][release-notes] for specific
 features and bug fixes.
 
+**NOTE: Deploying CoreOS on DigitalOcean is by standard not secure,** because DigitalOcean does not allow private networking. That means that even though you use private networking, you can still be attacked from others inside DigitalOceans private network.
+
 The following command will create a single droplet. For more details, check out
 <a href="#via-the-api">Launching via the API</a>.
 


os/booting-on-digitalocean.md
```diff
@@ -7,6 +7,8 @@ schedules per channel. You can [disable this feature][reboot-docs], although we
 don't recommend it. Read the [release notes][release-notes] for specific
 features and bug fixes.
 
+**NOTE: Deploying CoreOS on DigitalOcean is by standard not secure,** because DigitalOcean does not allow private networking. That means that even though you use private networking, you can still be attacked from others inside DigitalOceans private network.
```
robszumski: I wonder if it makes more sense to put this down on 110 near this, since it touches on the "private networking":

> The `$private_ipv4` and `$public_ipv4` substitution variables are fully
supported in cloud-config on DigitalOcean. In order for `$private_ipv4` to be
populated, the droplet must have private networking enabled.

It'd be great to work this into a sentence without having to say "note:". What do you think of this?

> The `$private_ipv4` and `$public_ipv4` substitution variables are fully
supported in cloud-config on DigitalOcean. In order for `$private_ipv4` to be
populated, the droplet must have "private networking" enabled. Private networking isn't available in all regions and isn't actually private from others; instead you can think of it as faster, free bandwidth between your droplets. etcd should be [configured with TLS](https://coreos.com/etcd/docs/latest/security.html#example-3-transport-security-&-client-certificates-in-a-cluster) when using this network.

kevinsimper: Yes, that is good, but the last line is more like:
**to prevent others from ~~reading your traffic~~ controlling your cluster**.

robszumski: Sounds good. Can you modify the PR accordingly.

One note: I added the bold just to show what was new, I don't think we need to bold the whole thing :)

joshix:
@kevinsimper Thanks for the excellent suggestion to add this note. I've adapted a version based on your and robszumski's input and am closing this in favor of those changes in #728.

PR #727, opened 2016-01-13
Added "trap" into docker "while" loop examples

kayrus:
When you don't use `trap`, `bash`/`sh` doesn't catch TERM signal which `docker stop` sends.
This causes `docker stop` timeout and then `docker kill`.

/cc @robszumski @joshix 

steps to reproduce with the Ubuntu container + bash (this example doesn't handle INT):

* `docker run -ti --rm ubuntu /bin/bash -c "while true; do echo hello world; sleep 1; done"`
* try to type `Ctrl+C`

or this busybox example doesn't handle TERM:

* `docker run -ti --name busybox --rm busybox /bin/sh -c "while true; do echo hello world; sleep 1; done"`
* Then in another window to run `docker stop busybox` it will timeout, then docker will kill it.

Both examples with the `trap 'exit 0' INT TERM;` work fine.


joshix:
This makes sense and LGTM.

PR #728, opened 2016-01-14
os/booting-on-digitalocean: Warn about "private" networking

### LINKED ISSUE #718: Insecure private networking with DigitalOcean

joshix:
Adopt @robszumski's version of changes suggested by @kevinsimper
to warn digital ocean users about implications of "private"
networking in the DO context.

Supersedes #724. Fixes #718.

### DIFF
--- a/os/booting-on-digitalocean.md
+++ b/os/booting-on-digitalocean.md
@@ -110,12 +110,19 @@ coreos:
       command: start
 ```
 
-The `$private_ipv4` and `$public_ipv4` substitution variables are fully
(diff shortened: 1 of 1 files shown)


robszumski:
LGTM

PR #730, opened 2016-01-19
os/: Fix links that were externally declared

joshix:
Repair markdown links that were using the removed external
reference links jekyll plugin.

Related to ~~and should merge along with~~ coreos-inc/coreos-pages#440.

(not necessary to merge simultaneously with the related, larger changes in the existing site (440), but the two should go fairly close together, logically)


robszumski:
LGTM

joshix:
interesting sample of our version scale: 6 source files, 246 files in the -pages repo

PR #734, opened 2016-01-26
Added doc on how to reconfigure failed etcd member in CoreOS

kayrus:
/cc @joshix @robszumski 

First draft.


etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
+# Restore etcd Cluster on CoreOS
+
+Here you will find step-by-step instructions on how to recover your etcd cluster. It is important to know that you can not restore your etcd cluster using only discovery URL. Discovery URL is used only once at bootstrap.
+
+In this example we will review 3-members etcd cluster with one failed node. It is important to take into consideration that this example supposes you used [Cloud-Config][cloud-config] with [discovery URL][etcd-discovery] to bootstrap your cluster with the following default options:
+
+```yaml
```
robszumski: Can we add `#cloud-config` for consistency with the rest of our snippets?

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+  etcd2:
+    advertise-client-urls: http://%PEER_IP_ADDRESS%:2379
```
robszumski: @joshix What do we do for input like this? For example, I've done `http://<PEER_IP_ADDRESS>:2379` in the past.

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+    initial-advertise-peer-urls: http://%PEER_IP_ADDRESS%:2380
+    listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
+    listen-peer-urls: http://0.0.0.0:2380
+    discovery: %DISCOVERY_URL%
+```
+
+If you have etcd cluster with TLS, just use `https://` instead of `http://` in commands' examples below. Make sure you didn't set custom etcd [member name][etcd-member-name] (by default CoreOS uses [machine-id] for etcd member name). Let's assume that your etcd cluster has faulty `172.16.0.102` member:
```
robszumski: machine id link is missing the reference

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
+# Restore etcd Cluster on CoreOS
+
+Here you will find step-by-step instructions on how to recover your etcd cluster. It is important to know that you can not restore your etcd cluster using only discovery URL. Discovery URL is used only once at bootstrap.
+
+In this example we will review 3-members etcd cluster with one failed node. It is important to take into consideration that this example supposes you used [Cloud-Config][cloud-config] with [discovery URL][etcd-discovery] to bootstrap your cluster with the following default options:
```
robszumski: "assumes" instead of "supposes"

etcd/restore-etcd-cluster-on-coreos.md
robszumski: why do we care if they set a custom name for the member? if its important, we need to explain why.

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+```sh
+$ etcdctl cluster-health
+member fe2f75dd51fa5ff is healthy: got healthy result from http://172.16.0.101:2379
+failed to check the health of member 1609b5a3a078c227 on http://172.16.0.102:2379: Get http://172.16.0.102:2379/health: dial tcp 192.168.122.69:2379: connection refused
+member 1609b5a3a078c227 is unreachable: [http://172.16.0.102:2379] are all unreachable
+member 60e8a32b09dc91f1 is healthy: got healthy result from http://172.16.0.103:2379
+cluster is healthy
+```
+
+First of all `etcdctl` on failed node should be configured to use healthy endpoint, so we have to configure environment variable and you will be able to run `etcdctl` command on your failed node:
```
robszumski: Can we sidestep 27-31 if we just say, "run etcdctl from a working node"?

This also brings up a bigger question, what type of failure are we trying to recover from here? If you can get on the node, I would imagine that is a disk failure or similar. We should probably state this at the top.

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+
+```sh
+export ETCDCTL_ENDPOINT=http://172.16.0.101:2379
+```
+
+Then we have to remove failed 172.16.0.102 member from your etcd cluster:
```
robszumski: should we explain what this command will do, or link to the admin guide? "this will tell all the other nodes that a human has determined this node is dead and not to listen to it ever again"

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+$ sudo systemd-tmpfiles --create /usr/lib64/tmpfiles.d/etcd2.conf
+```
+
+or with this command:
+
+```sh
+$ sudo install -m 755 -d -o etcd -g etcd /var/lib/etcd2
+```
+
+By default etcd uses [machine-id] for the etcd member name. Then we have to reinitialize failed member with this machine-id:
```
robszumski: When reading this top to bottom, I had _no idea_ that we were trying to recover this node, not just remove it and shut it down. We need to clearly state that at the top. The only hint is the title.

robszumski: "using the machine id is useful because it is already a uuid, which is globally unique. other coreos systems also reference machines by this id, such as locksmith."

etcd/restore-etcd-cluster-on-coreos.md
```diff
@@ -0,0 +1,110 @@
...
+$ sudo systemctl start etcd2
+```
+
+You can check whether your cluster is healthy:
+
+```sh
+$ etcdctl cluster-health
+```
+
+When your cluster has healthy state that means that etcd successfully wrote cluster configuration into `/var/lib/etcd2` directory. And now it is safe to remove `/run/systemd/system/etcd2.service.d/99-restore.conf` drop-in. Or leave it as is because on next boot it will be cleaned up automatically.
```
robszumski: Don't you need to keep this since cloudinit will still write out its dropin?

etcd/restore-etcd-cluster-on-coreos.md
joshix: `http://<PEER_ADDRESS>:2379` or `http://$PEER_ADDRESS:2379` (depending on the context, using a bash variable style like the second form may make the most sense).

Not `%PEER_ADDRESS%` as in this line -- let's not add a third variable form.

etcd/restore-etcd-cluster-on-coreos.md
kayrus: it is not uuid, just random 32bit hash

etcd/restore-etcd-cluster-on-coreos.md
kayrus: If you use same member name it is not necessary. Added a note above.

etcd/restore-etcd-cluster-on-coreos.md
kayrus: done

etcd/restore-etcd-cluster-on-coreos.md
kayrus: done. added a note at the top

etcd/restore-etcd-cluster-on-coreos.md
kayrus: done

etcd/restore-etcd-cluster-on-coreos.md
kayrus: If you use same member name it is not necessary. Added a note.

etcd/restore-etcd-cluster-on-coreos.md
kayrus: I suppose `$PEER_ADDRESS` can confuse users that this variable could be substituted. Let's use <PEER_ADDRESS>.

etcd/restore-etcd-cluster-on-coreos.md
kayrus: also added `coreos:`

robszumski:
Note for the future: can you only rebase at the end so we can see the changes happen per commit?

os/using-environment-variables-in-systemd-units.md
```diff
@@ -87,51 +87,7 @@ This unit file will run nginx Docker container and bind it to specific IP addres
...
-```
-
-Run `sudo systemctl daemon-reload` and `sudo systemctl restart etcd2` to apply your changes. You will see that your proxy node became cluster member:
-
-```
-etcdserver: start member 9bf1b35fc7761a23 in cluster 36cce781cb4f1292
-```
-
-Once your proxy node became member node and `etcdctl cluster-health` shows healthy cluster, you can remove your temporarily drop-in `sudo rm /run/systemd/system/etcd2.service.d/99-restore.conf && sudo systemctl daemon-reload`.
+This example moved [here][etcd-cluster-reconfiguration].
```
robszumski: Can we phrase this differently? A new user will never know that this was here.

robszumski: It would also be more informative to add a description of what the example shows. What about:

"A complete example of combining environment variables and systemd drop-ins to reconfigure an existing machine running etcd."

etcd/etcd-cluster-runtime-reconfiguration-on-coreos.md
```diff
@@ -0,0 +1,187 @@
+# etcd Cluster Runtime Reconfiguration on CoreOS
+
```
robszumski: We still need a really high level overview here. Probably the same comment from https://github.com/coreos/docs/pull/734/files#r51036485 or similar.

kayrus: If we add systemd and remove CoreOS - this doc will not reflect the truth.

os/using-environment-variables-in-systemd-units.md
kayrus: done in next commit

etcd/etcd-cluster-runtime-reconfiguration-on-coreos.md
robszumski: ok, we just need to put whatever describes it.

"Use this document to reconfigure or recover an etcd cluster running on CoreOS, using a combination of systemd features and etcdctl commands."

kayrus: Done

joshix:
Could we rename this file to `etcd-live-cluster-reconfiguration.md`? -- to be shorter and to match the organic convention of file names like `etcd-live-http-to-https-migration.md`

kayrus:
@joshix done

joshix:
@kayrus LGTM. Please squash for merge. Thanks!

joshix:
(decided we could forego the squash in favor of going ahead and getting this in)

PR #738, opened 2016-01-29
Define drop-ins in proper way

kayrus:
/cc @robszumski @joshix

### DIFF
--- a/os/customizing-docker.md
+++ b/os/customizing-docker.md
@@ -309,17 +309,14 @@ The easiest way to use these new ulimits on all of your machines is via cloud-co
 ```yaml
 #cloud-config
 
-write_files:
-    - path: /etc/systemd/system/docker.service.d/increase-ulimit.conf
-      owner: core:core
-      permissions: 0644
-      content: |
-        [Service]
-        LimitMEMLOCK=infinity
-
 coreos:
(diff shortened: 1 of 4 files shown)


os/customizing-docker.md
```diff
@@ -309,17 +309,14 @@ The easiest way to use these new ulimits on all of your machines is via cloud-co
 ```yaml
 #cloud-config
 
-write_files:
-    - path: /etc/systemd/system/docker.service.d/increase-ulimit.conf
-      owner: core:core
-      permissions: 0644
```
robszumski: we're losing these permissions, does that matter?

joshix: @robszumski I *think* that the units will get default and appropriate permissions when created in this fashion (specified in terms of their parent unit). Before this change, we needed to specify the file mode because we were creating it with a brute force `write_files:`. Now we don't. @kayrus can you confirm if I'm correct?

crawford: Correct. Losing the file permissions is fine.

crawford:
LGTM

os/customizing-docker.md
kayrus: Agree with @crawford

Issue #742, opened 2016-02-04
Rackspace provisioning guide is borked


iangcarroll:
https://coreos.com/os/docs/latest/booting-on-rackspace.html

Looks like a div didn't get closed, as the rest of the page is under the "Alpha" tab.

robszumski:
This was caused by a switch to the markdown renderer for coreos.com. Thanks for reporting, we're looking into it.

For anyone interested, we're using Jekyll 3 and were encouraged by Github's switch to Kramdown (we're using rdiscount currently). It has trouble rendering mixed HTML and markdown. This is part of the markdown "spec", however.

crawford:
@robszumski that page has unclosed anchor tags.

PR #743, opened 2016-02-05
os: add missing closing anchor tags

crawford:
Fixes https://github.com/coreos/docs/issues/742.


marineam:
Makes you wonder if there are things to validate HTML.... ;-) LGTM

crawford:
@marineam how do you think I tracked down this one? :)

PR #747, opened 2016-02-08
os: correct external storage config

crawford:
The mount unit didn't Require/Wants docker.service, so the After clause has no
effect. It's a bit more robust to use the drop-in anyway.

Fixes https://github.com/coreos/bugs/issues/1088.


robszumski:
LGTM

PR #762, opened 2016-02-23
os: clean up cloudstack docs

crawford:
Let's use markdown instead of HTML.


robszumski:
Weird, how'd that get in there. LGTM.

Issue #755, opened 2016-02-19
os doc is out of date.


xiang90:
https://coreos.com/os/docs/latest/cluster-discovery.html

The https://coreos.com/os/docs/latest/cluster-discovery.html#existing-clusters section is not correct for etcd2 cluster.

We should replace it with https://github.com/coreos/etcd/blob/master/Documentation/runtime-reconf-design.md#do-not-use-public-discovery-service-for-runtime-reconfiguration

robszumski:
dup of https://github.com/coreos/docs/issues/739

I do think there is value in explaining how the system works, so we should update the existing doc if it makes sense.

xiang90:
@robszumski I agree. As I suggested, we probably should just replace that section with the new one I listed.

rca:
I found this issue via [etcd #4568](https://github.com/coreos/etcd/issues/4568), and while updating the docs is a great idea, they still do not describe what Issue #4568 asks:  How does one add to the cluster when a discovery URL has been used?

I have the same question, and [asked it on the Google Group](https://groups.google.com/forum/#!topic/coreos-user/DTmUyBiqoO0).  Instructions on doing so would be great.

In my case, I have an existing 7-node EC2 cluster where all nodes were created using public discovery.  I want to replace one of the cluster members with one of the other 4 proxy nodes.  What is the process for this?  Given the process, I'd be more than happy to update the docs.

Thank you!

joshix:
@rca In short, one should switch to static discovery and then promote a proxy. That procedure(s) for etcd on CoreOS is described in [etcd cluster reconfiguration](https://github.com/coreos/docs/blob/master/etcd/etcd-live-cluster-reconfiguration.md). Feedback and or PRs on that document as you apply it in your environment are also welcome.

If you've deployed on another OS, the same changes need to be made but without the cloud-config facility (and potentially adapted for your init system if you're not on planet systemd.)

Let me know if that is helpful or if you run into any problems.

rca:
The doc you pointed me to is pretty much what I was looking for, thanks, @joshix!

[...]

Issue #739, opened 2016-01-31
Cluster discovery docs are out of date


crawford:
https://coreos.com/os/docs/latest/cluster-discovery.html

This is not how etcd2 works. I'm not sure why these docs are under the OS section either.

robszumski:
These are under the OS section because there wasn't an upstream doc at the time, plus it is OS-specific as it covers cloud-init, etc.

It can be removed.

PR #771, opened 2016-03-08
os/cluster-discovery: Update cluster reconfiguration info for later etcd

joshix:
Stop recommending using the public discovery service for anything but
bootstrap. Don't reconfigure a cluster with public discovery.

Fixes coreos/docs#755, fixes coreos/etcd#4568, fixes coreos/docs#739.


gyuho:
lgtm. I will defer to @xiang90 and @crawford. Thanks!

xiang90:
lgtm. Thanks!

PR #774, opened 2016-03-14
update-strategies: use mask and stop

philips:
The best way of accomplishing this is to mask and stop. Also, add docs for people who may wish to do this manually.

### DIFF
--- a/os/update-strategies.md
+++ b/os/update-strategies.md
@@ -82,9 +82,15 @@ coreos:
       command: stop
     - name: locksmithd.service
       command: stop
+    - name: update-engine.service
+      command: mask
+    - name: locksmithd.service
+      command: mask
 ```
 
-## Updating behind a proxy
+Alternatively, to manually mask update-engine run `systemctl mask update-engine.service`. To reverse this remove the stanza from the cloud-config and run `systemctl unmask update-engine.service`
+
+## Updating Behind a Proxy
 
 Public Internet access is required to contact CoreUpdate and download new versions of CoreOS. If direct access is not available the `update-engine` service may be configured to use a HTTP or SOCKS proxy using curl-compatible environment variables, such as `HTTPS_PROXY` or `ALL_PROXY`.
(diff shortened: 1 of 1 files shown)


philips:
cc @crawford

os/update-strategies.md
```diff
@@ -89,8 +89,14 @@ coreos:
       command: stop
     - name: locksmithd.service
       command: stop
+    - name: update-engine.service
+      command: mask
+    - name: locksmithd.service
+      command: mask
 ```
 
+Alternatively, to manually mask update-engine run `systemctl mask update-engine.service`. To reverse this remove the stanza from the cloud-config and run `systemctl unmask update-engine.service`
```
crawford: "remove the stanza from the cloud-config" This is not possible on many platforms. We should stop suggesting that cloud-init be used for configuration management. It's really only good for static configuration.

os/update-strategies.md
```diff
@@ -89,8 +89,14 @@ coreos:
       command: stop
     - name: locksmithd.service
       command: stop
+    - name: update-engine.service
+      command: mask
+    - name: locksmithd.service
+      command: mask
```
crawford: Why mask it? cloud-init already stops it.

os/update-strategies.md
```diff
@@ -89,8 +89,14 @@ coreos:
       command: stop
     - name: locksmithd.service
       command: stop
+    - name: update-engine.service
+      command: mask
```
crawford: If this is needed, we should add it to cloud-init.

ElijahCaine:
@philips I'm cleaning up some old PRs.  Any thoughts on this?  If you think it should be merged we should address @crawford's feedback, otherwise I can close this.

crawford:
This is probably fine. coreos-cloudinit is garbage and we can tell users to switch to Ignition when they come to the same realization.

ElijahCaine:
I rebased this PR to get rid of the conflicts.
@philips @joshix Think this should be merged after @crawford's last comment?

PR #779, opened 2016-03-25
Update instructions for newer version of VMWare Fusion

YogoGit:
As a newbie to CoreOS (but a 30 year veteran to unix and a 2 year Docker user), the instructions were somewhat confusing when trying to spinup CoreOS on VMware Fusino 8.1 (the latest version).  After playing around for a few hours, I got it working.  This is an attempt to make it easier for users in the same situation.

### DIFF
--- a/os/booting-on-vmware.md
+++ b/os/booting-on-vmware.md
@@ -38,10 +38,6 @@ CoreOS is released into stable, alpha and beta channels. Releases to each channe
 
 [release notes]: https://coreos.com/releases/
 
-### Booting with VMware Fusion
-
-After downloading the VMware OVA image for your selected channel, open the `coreos_production_vmware_ova.ova` file to create a VM.
-
 ### Booting with VMware ESXi
 
 Use the vSphere Client to deploy the VM as follows:
@@ -59,7 +55,7 @@ The last step uploads the files to the ESXi datastore and registers the new VM. 
 
 *NB: These instructions were tested with an ESXi v5.5 host.*
 
-### Booting with VMware Workstation 12
+### Booting with VMware Workstation 12 or VMware Fusion
 
 Run VMware Workstation GUI:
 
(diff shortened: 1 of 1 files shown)


os/booting-on-vmware.md
```diff
@@ -68,7 +64,12 @@ Run VMware Workstation GUI:
 3. Name your VM, then click `Import`
 4. (Press `Retry` *if* VMware Workstation raises an "OVF specification" warning)
 5. Edit VM settings if necessary
-6. Start your CoreOS VM
+6. Create a cloud-config image (see below) containing at least at least one valid SSH key using the instructions at:
```
dghubble: at least at least

robszumski: Can we find a way to link some text instead of the raw link? Like:

...one valid SSH key using [a config-drive](https://github.com/coreos/coreos-cloudinit/blob/master/Documentation/config-drive.md).

os/booting-on-vmware.md
```diff
@@ -68,7 +64,12 @@ Run VMware Workstation GUI:
...
+    https://github.com/coreos/coreos-cloudinit/blob/master/Documentation/config-drive.md
+    * On Fusion, simply connect and mount this image as the CD, ignoring the instruction below regarding the filesystem label which are not applicable.
```
robszumski: Maybe make it clear that you are attaching the config drive? I think that repeating that phrase here will work well:

...simply connect and mount this config-drive image as the CD

os/booting-on-vmware.md
YogoGit: Fixed the double double line and inlined the raw link

os/booting-on-vmware.md
YogoGit: It wasn't obvious to me, which is why I wasted time trying to follow the instructions, since it implies that you connect to the drive, but there isn't any way to provide a label on mounted CD drives on Fusion (perhaps it's possible in VMware Workstation and/or ESXi).

robszumski:
LGTM, thanks!

PR #787, opened 2016-04-11
os: fix instructions for customizing sshd

philips:
I tested this and it works for me on 1010.1.0 and seems to have been the
intention based on the discussion here: https://github.com/coreos/bugs/issues/426

cc @joshix @crawford


githubuseracct:
+1. Tested on stable 899.15.0. The combination of both options is the only thing that worked reliably for me.

robszumski:
LGTM, I was trying to track down what we wanted to document here as well.

PR #792, opened 2016-04-18
Add initial attempt at iSCSI documentation

mjg59:


robszumski:
Great start, thanks for putting this together. 

@joshix Brainstorming some steps to polish:

- Background
  - add some general background on how iscsi works, just a sentence or two. It's a block device, etc.
- Credentials section has examples for global and per-mount
  - manual examples
  - cloud-configs (enables iscsid)
- Expand on the "find available targets"
  - "You may want to manually search first, then add to a cloud-config..."
- Mounting instructions
  - doc has a specific iSCSI example (currently is lacking)
  - full cloud-config
  - possibly add a kubernetes example? ([upstream](https://github.com/kubernetes/kubernetes/blob/master/examples/iscsi/iscsi.json))

mjg59:
@robszumski Ah yeah things get a little more complicated if you don't want to do discovery on the machine - the configuration for the target needs to be laid down in that case before login will work. We'll need to provide an example for that.

ElijahCaine:
Closed in favor of #890

PR #797, opened 2016-04-20
qe/geo-replication.md: Clarify replicated pull paragraph

joshix:


joshix:
Solo merge as it is a follow-up edit to previous PR

PR #798, opened 2016-04-25
etcd/: Update 10 rel links to crawfred style

joshix:
Match downstream backport prequel coreos-inc/coreos-pages#639


crawford:
This is not correct. All links within this repo need to resolve. Please revert.

joshix:
When do we do the rewrite?

crawford:
The import script handles all rewrites.

joshix:
It didn't handle these -- I did them by hand in coreos-inc/coreos-pages#639

crawford:
We need to fix the script then.

joshix:
I squashed that commit away, but I had to go through and replace `../os/` with `../../../os/docs/latest/` throughout _etcd/2.3.2/ and _etcd/latest/ -- *in the etcd docs that live in coreos/docs/etcd/*

PR #804, opened 2016-05-02
ignition: add boot-process document

crawford:


ignition/boot-process.md
```diff
@@ -0,0 +1,37 @@
+# CoreOS startup process
+
+The CoreOS startup process is fundamentally built on the standard [Linux startup process][linux startup]. Since this process is already well documented and generally well understood, this document will focus on particular aspects specific to CoreOS's startup process.
```
joshix: s/fundamentally//
s/particular//
"specific to booting CoreOS."

ignition/boot-process.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Bootloader
+
+The first component of CoreOS to execute on a booting system is [GRUB][grub]. The [GRUB config][grub config] used by CoreOS serves several important roles.
```
joshix: `[GRUB][grub] is the first program executed when a CoreOS system boots.`

ignition/boot-process.md
```diff
@@ -0,0 +1,37 @@
...
+
+First off, it [determines which USR partition to use][gptprio.next] which is what allows CoreOS to do its automatic, atomic upgrades and rollbacks.
```
joshix:
```
The CoreOS [GRUB config][grubconfig] has several roles. First, the GRUB config [specifies which `usr` partition to use][gptprio.next] from the two `usr` partitions CoreOS uses to provide atomic upgrades and rollbacks.
```

ignition/boot-process.md
joshix: Strike second sentence; replaced in line 9 below

crawford:
Updated. This doc is a little half-assed right now. I think it's fine in regard to Ignition, but I think detailing a lot more of the OS would be helpful. And if we went that route, I don't think this would live in the Ignition section.

ignition/boot-process.md
```diff
@@ -0,0 +1,37 @@
...
+
+The CoreOS [GRUB config][grubconfig] has several roles. First, the GRUB config [specifies which `usr` partition to use][gptprio.next] from the two `usr` partitions CoreOS uses to provide atomic upgrades and rollbacks.
+
+Additionally, GRUB determines if this is the first time a machine has booted. This is accomplished by [searching for the initial disk GUID][search] (00000000-0000-0000-0000-000000000001). CoreOS is built with this well-known disk GUID in order to detect the first boot. This GUID is randomized later in the boot process so that individual disks may be uniquely identified. If GRUB detects that this is in fact a first boot, it sets two Linux kernel command line parameters: `coreos.randomize_guid=00000000-0000-0000-0000-000000000001` and `coreos.first_boot=1`. These parameters are used by various programs in later stages of the boot process.
+
+## Early user space
+
+The next major milestone in the CoreOS startup process is the jump into the initial RAM file system. In addition to its standard responsibility of mounting the root filesystem, the initramfs is also where the disk GUID is randomized and where Ignition runs.
```
robszumski: Probably need to work in "sometimes referred to as the initrd" somewhere

crawford: The initrd and initramfs are actually different things. They serve the same purpose, but their semantics and format are different (initrd is legacy). If people recognize "initrd", I'm pretty sure they will recognize "initramfs" as well. Maybe not though.

joshix:
LGTM

PR #805, opened 2016-05-02
Add outline SSSD documentation

mjg59:


mjg59:
This probably needs significantly more content, but it's a starting point

joshix:
(Merged for pairing with experimental SSSD support in latest alpha OS release; expansion later)

PR #813, opened 2016-05-06
os: remove rackspace ids and broken image

crawford:


joshix:
LGTM

joshix:
ref coreos-inc/coreos-pages#659

PR #814, opened 2016-05-06
Fix docs

josephschorr:


joshix:
Also needs a fix downstream in tectonic-pages/_quay-enterprise/docs/latest/clair.md

joshix:
based on what worked downstream when pre-back-porting this, ;), LGTM

PR #820, opened 2016-05-13
Put initrd on separate line to support multiple initramfs

schlomo:
With this change users don't have to extract the CoreOS image but can simply add their additions as a second initramfs. Much less effort.

See http://www.syslinux.org/wiki/index.php?title=SYSLINUX#INITRD_initrd_file for background info.


joshix:
Sorry this took so long to give a minimal test. Thanks for the contrib!

PR #824, opened 2016-05-18
etcd/getting-started: Fix API button link

joshix:
Ref coreos-inc/coreos-pages#695


joshix:
(this will need to be updated in the etcd v3 docs update)

PR #826, opened 2016-05-18
update-strategies: remove mention of best-effort

philips:
best-effort is troublesome for a variety of resasons. We shouldn't be recommending it.

cc @mischief


mischief:
good riddance. lgtm

crawford:
YES

PR #827, opened 2016-05-19
os: added note to manual update when disabling update-engine

jbkc85:
os/update-strategies.md: Added a note to warn of potential issues in a manual update upon disabling update-engine.service after line #122

After experiencing issues after reboot noted in the issue coreos/bugs#1189, I had time to sit down and try to diagnose the issue.  I brought this to the attention of Brian Harrington as OSCON, he seemed to confirm the suspicion that with update-engine disabled upon reboot, the updates would eventually revert back to the passive partition after the following reboot.

IE:
update installed manually, reboot, see version NEW.x, generic reboot (for no specific reason), see version OLD.x

I was hoping to add this documentation to inform anyone else that are falling into the same trap.  If merged, I hope to add a Systemd Service File that includes an example of the 'OnCalendar' as mentioned above to shut of update-engine/locksmithd on a designated time after reboot to avoid issues like this in the future.

Thanks,
Jason Cameron

### DIFF
--- a/os/update-strategies.md
+++ b/os/update-strategies.md
@@ -119,6 +119,19 @@ $ update_engine_client -check_for_update
 [0123/220706:INFO:update_engine_client.cc(245)] Initiating update check and install.
 ```
 
+### Running Manual Update with update-engine disabled
+
+If you disabled update-engine.service and wish to run the manual update, you must first enable the update-service to run the update_engine_client:
+
+```
+$ systemctl start update-engine.service
+$ update_engine_client -check_for_update
+```
+
+After the update you must comment out or remove the disabling of the update-engine.service for this first reboot. This allows for the update-engine to mark the new active partition, being your updated one, as a successful boot. If you do not run update-engine upon the first reboot, there is a chance that your machine will mark the new active partition as a failure and revert back to the passive partition (being the previous version) on the next reboot of the machine. Once rebooted you are able to again stop the update-engine and uncomment or add the update-engine.service disabling in your cloud-config.yml.
+
+To avoid this process, a potential option is to use a Service for disabling update-engine and locksmithd using the [Systemd OnCalendar setting](oncalendar-option) which will disable the update-engine.service and locksmithd.service in a particular time period after reboot.
+
 ## Auto-updates with a maintenance window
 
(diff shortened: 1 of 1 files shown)


edevil:
Is this the "official" way of running a system with manual updates? I've also ran into this problem of reverting to a previous version after a normal reboot.

crawford:
@jbkc85 Thanks for the pull request and sorry for the ridiculous delay. Because of the degree to which update-engine is tied into the reboot logic, I cannot recommend it ever be stopped (contrary to some of our out-of-date documentation). This means that locksmith is going to be the utility that controls when your machine reboots. This can be used to set up maintenance windows, coordinated reboots, etc. and this is what we should focus on in our docs.

@edevil No. The official way to achieve "manual updates" is to operate your own [CoreUpdate](https://coreos.com/products/coreupdate) and choose the specific version of Container Linux that you'd like your machines to use.

jbkc85:
You act like you guys over there at CoreOS are busy!

So if I understand correctly, the documentation should really reflect the following in concerns to "manual updates":

1. Set up CoreUpdate server
2. Configure boxes (either cloud-config.yml, ignition or ``/etc/coreos/update.conf``) for new update server
3. Control updates via CoreUpdate and Locksmith

I know that there are a lot of in between steps that would need to be documented, but is this the route that would be preferred over the current documented disable update-engine?

crawford:
That's correct. We will definitely be leading the charge there, I don't expect you to do all that for us. I just wanted to give you a decent explanation for why I won't merge your pull request.

jbkc85:
Awesome, I am just glad I am on the same page!

As for the documentation, I understand you guys are leading the charge, but will definitely be trying this on premise with my own cluster and sharing the docs/results from it.  CoreOS has given much to the community and if this is a way I can give back I will try my best!

jbkc85:
@crawford I didn't realize that CoreUpdate was part of the Managed Linux solution CoreOS is offering.  I don't suppose there is an equivalent in the free-to-use realm is there?  If not, I understand - but is there any way to get a trial on the CoreUpdate so I could help write up some documentation (if needed!)

crawford:
There is not anything open-source, but I've had my eyes on that for a while. I'd like to split off a chunk of CoreUpdate into a basic, open-source tool. In the mean time, I think we are good on the documentation for the existing tools. Thanks for the offer!

PR #833, opened 2016-06-10
launching-containers-fleet.md: typo

petrus-v:
*select* with an `s` as first character


petrus-v:
My bad elect is better here!

PR #834, opened 2016-06-13
Update registry-authentication.md

burningTyger:
Sorry if this doesn't apply to CoreOS but the official Docker client stores data in this file and this page turn up when you search for docker login credentials.

PR #835, opened 2016-06-13
os: make file path clear in registry auth

robszumski:
Clear up the file path from https://github.com/coreos/docs/pull/834


crawford:
LGTM

PR #844, opened 2016-06-23
os/network-config-with-networkd.md - Fix heading for Ignition debug option.

ErnestG4:
Ignition debug drop-in was mislabeled with a Cloud-Config heading. Changed to the correct "Enable debugging through Ignition".


crawford:
LGTM

Issue #843, opened 2016-06-22
Docker storage instructions incomplete/incorrect


wintamute:
The instructions on how to add ephemeral [storage to Docker] (https://github.com/coreos/docs/blob/master/os/mounting-storage.md#use-attached-storage-for-docker) are missing a critical piece of information which leads to incorrect behavior.

```yaml
  - name: format-ephemeral.service
       command: start
       content: |
         [Unit]
         Description=Formats the ephemeral drive
         After=dev-xvdb.device
         Requires=dev-xvdb.device
         [Service]
         Type=oneshot
         RemainAfterExit=yes
         ExecStart=/usr/sbin/wipefs -f /dev/xvdb
         ExecStart=/usr/sbin/mkfs.ext4 -F /dev/xvdb
```

correctly formats the drive, but does so **on every reboot** which kinda defeats the purpose of ephemeral storage.
Using the following works as intended (as suggested [here](https://groups.google.com/forum/#!topic/coreos-user/-OYyza6LVpk))
```yaml
 - name: format-ephemeral.service
       command: start
       content: |
         [Unit]
         Description=Formats the ephemeral drive
         After=dev-sdb.device
         Requires=dev-sdb.device
         ConditionPathExists=!/var/lib/format-done
         [Service]
         Type=oneshot
         RemainAfterExit=yes
         ExecStart=/usr/sbin/wipefs -f /dev/sdb
         ExecStart=/usr/sbin/mkfs.ext4 -F /dev/sdb
         ExecStartPost=/usr/bin/touch /var/lib/format-done
```

Using the also suggested option 
> ConditionFirstBoot=

might work too, but wasn't a solution for me since the disk was added after first booting.
Not sure what the preferred solution to this would be, but at least it should be pointed out that with the current Unit file you'll loose everything after the next reboot.

ElijahCaine:
Update: PR created, once that gets merged or closed this will be closed.

PR #846, opened 2016-06-24
os: prevent formatting from happening twice

robszumski:
Fixes https://github.com/coreos/docs/issues/843

cc: @wintamute

### DIFF
--- a/os/mounting-storage.md
+++ b/os/mounting-storage.md
@@ -22,7 +22,9 @@ coreos:
 
 Docker containers can be very large and debugging a build process makes it easy to accumulate hundreds of containers. It's advantageous to use attached storage to expand your capacity for container images. Be aware that some cloud providers treat certain disks as ephemeral and you will lose all Docker images contained on that disk.
 
-We're going to mount a ext4 device to `/var/lib/docker`, where Docker stores images. We can do this on the fly when the machines starts up with a oneshot unit that formats the drive and another one that runs afterwards to mount it. Be sure to hardcode the correct device or look for a device by label:
+We're going to mount a ext4 device to `/var/lib/docker`, where Docker stores images. We can do this on the fly when the machines starts up with a "oneshot" unit that formats the drive and another one that runs afterwards to mount it. To ensure our disk only gets formatted once, the unit will only run if the file `/var/lib/format-xvdb` isn't present. Upon successful formatting, we'll write out that file in an `ExecStopPost=` command.
+
+Be sure to hardcode the correct device or look for a device by label:
 
 ```yaml
 #cloud-config
@@ -35,11 +37,13 @@ coreos:
         Description=Formats the ephemeral drive
         After=dev-xvdb.device
         Requires=dev-xvdb.device
+        ConditionPathExists=!/var/lib/format-xvdb
         [Service]
         Type=oneshot
         RemainAfterExit=yes
         ExecStart=/usr/sbin/wipefs -f /dev/xvdb
         ExecStart=/usr/sbin/mkfs.ext4 -F /dev/xvdb
+        ExecStartPost=/usr/bin/touch /var/lib/format-xvdb
     - name: var-lib-docker.mount
       command: start
       content: |


os/mounting-storage.md
```diff
@@ -35,11 +37,13 @@ coreos:
         Description=Formats the ephemeral drive
         After=dev-xvdb.device
         Requires=dev-xvdb.device
+        ConditionPathExists=!/var/lib/format-xvdb
```
crawford: I would use ConditionFirstBoot instead

robszumski: This is specifically noted as a problem in the original issue https://github.com/coreos/docs/issues/843

crawford: I don't think we should support that use case. Adding storage after the fact is a very manual operation which goes against nearly everything we produce. We should instead document a fix for that user's root cause.

wintamute: Just some additional info: I'm running CoreOS on top of Openstack. When creating a new instance, it's afaik not possible to add a storage volume (which is independent of the instance lifetime) right away. So you can only add a volume after the instance has already booted. That's why I stated that in my case, ConditionFirstBoot didn't work for me. Of course the volume could be formatted outside and just attached.
My main concern related to this issue is that the current Unit file might lead to data loss due to re-formatting drives in use.

crawford: As long as you attach the device before the first reboot, this should be fine (though, I haven't actually tested it myself), so a `nova boot` followed by a `nova volume-attach` should work. That does feel a bit fragile though, so maybe a call to `blkid` to test for a formatted filesystem would be more robust.

robszumski:
I attempted to rewrite this with the following cloud-config, just can't get the all-on-one-line syntax correct in the `ExecStartPre`. Any pointers @crawford? systemctl status output below.

The tricky part is I want to invert the exit codes, such that finding a fs exits 1, but no fs exits 0, so the unit will continue to run/format.

```
#cloud-config
coreos:
  units:
    - name: format-ephemeral.service
      command: start
      content: |
        [Unit]
        Description=Formats the ephemeral drive if the desired filesystem does not exist
        After=dev-xvdb.device
        Requires=dev-xvdb.device
        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStartPre=/bin/bash -c "case "$(/usr/bin/lsblk -no FSTYPE /dev/xvdb)" in ext4) exit 1;; *) exit 0;; esac"
        ExecStart=/usr/sbin/wipefs -f /dev/xvdb
        ExecStart=/usr/sbin/mkfs.ext4 -F /dev/xvdb
    - name: var-lib-docker.mount
      command: start
      content: |
        [Unit]
        Description=Mount ephemeral to /var/lib/docker
        Wants=format-ephemeral.service
        After=format-ephemeral.service
        [Mount]
        What=/dev/xvdb
        Where=/var/lib/docker
        Type=ext4
    - name: docker.service
      drop-ins:
        - name: 10-wait-docker.conf
          content: |
            [Unit]
            After=var-lib-docker.mount
            Requires=var-lib-docker.mount
```

```
al bash[812]: -no: -c: line 0: unexpected EOF while looking for matching `)'
al bash[812]: -no: -c: line 1: syntax error: unexpected end of file
al systemd[1]: format-ephemeral.service: Control process exited, code=exited status=1
al systemd[1]: Failed to start Formats the ephemeral drive if the desired filesystem does not exist.
al systemd[1]: format-ephemeral.service: Unit entered failed state.
al systemd[1]: format-ephemeral.service: Failed with result 'exit-code'.
al systemd[1]: Starting Formats the ephemeral drive if the desired filesystem does not exist...
al bash[816]: -no: -c: line 0: unexpected EOF while looking for matching `)'
al bash[816]: -no: -c: line 1: syntax error: unexpected end of file
al systemd[1]: format-ephemeral.service: Control process exited, code=exited status=1
al systemd[1]: Failed to start Formats the ephemeral drive if the desired filesystem does not exist.
```

crawford:
You had some extra quotes in there.

```sh
/bin/bash -c "case $(lsblk -no FSTYPE /dev/xvdb) in ext4) exit 1;; *) exit 0;; esac"
```

robszumski:
Had a few more seconds to look into this.

Here's what I ended up with:

```
#cloud-config
coreos:
  units:
    - name: format-ephemeral.service
      command: start
      content: |
        [Unit]
        Description=Formats the ephemeral drive if the desired filesystem does not exist
        After=dev-xvdb.device
        Requires=dev-xvdb.device
        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStartPre=/bin/bash -c 'case "$(lsblk -no FSTYPE /dev/xvdb)" in ext4) exit 1;; *) exit 0;; esac'
        ExecStart=/usr/sbin/wipefs -f /dev/xvdb
        ExecStart=/usr/sbin/mkfs.ext4 -F /dev/xvdb
    - name: var-lib-docker.mount
      command: start
      content: |
        [Unit]
        Description=Mount ephemeral to /var/lib/docker
        Wants=format-ephemeral.service
        After=format-ephemeral.service
        [Mount]
        What=/dev/xvdb
        Where=/var/lib/docker
        Type=ext4
    - name: docker.service
      drop-ins:
        - name: 10-wait-docker.conf
          content: |
            [Unit]
            After=var-lib-docker.mount
            Requires=var-lib-docker.mount
```

It works well and doesn't double format, and the unit does fail. It's not ideal that you'll see this every time you login though...

```
l systemd[1]: Starting Formats the ephemeral drive if the desired filesystem does not exist...
l systemd[1]: format-ephemeral.service: Control process exited, code=exited status=1
l systemd[1]: Failed to start Formats the ephemeral drive if the desired filesystem does not exist.
l systemd[1]: format-ephemeral.service: Unit entered failed state.
l systemd[1]: format-ephemeral.service: Failed with result 'exit-code'.
l systemd[1]: Starting Formats the ephemeral drive if the desired filesystem does not exist...
l systemd[1]: format-ephemeral.service: Control process exited, code=exited status=1
l systemd[1]: Failed to start Formats the ephemeral drive if the desired filesystem does not exist.
l systemd[1]: format-ephemeral.service: Unit entered failed state.
l systemd[1]: format-ephemeral.service: Failed with result 'exit-code'.
l systemd[1]: Starting Formats the ephemeral drive if the desired filesystem does not exist...
l systemd[1]: format-ephemeral.service: Control process exited, code=exited status=1
l systemd[1]: Failed to start Formats the ephemeral drive if the desired filesystem does not exist.
l systemd[1]: format-ephemeral.service: Unit entered failed state.
l systemd[1]: format-ephemeral.service: Failed with result 'exit-code'.
```

crawford:
@robszumski you could put everything in the ExecStart and then just no-op if it's already formatted. This is getting pretty gross.

ElijahCaine:
@robszumski @crawford What's that status of this document?

crawford:
The proper solution is to let Ignition format the filesystem instead. We can probably just yank this example entirely.

PR #847, opened 2016-06-25
harmonize client/server/peer concept naming

zined:
reading through the `Generate self-signed certificates` documentation, the concept of `client-server` certificates was clear after thinking about it for a while, but got easier to grasp and more consistent once I've replaced `client-server` by `peer` in my mind. As the concept `peer` is used throughout `etcd` already, I suggest adopting it here, too.


ElijahCaine:
@zined This looks good to me, I don't think there's any conflicts with moving `client-server` to `peer`, and consistency is always good.
@joshix any thoughts on this change?  Unless it was called `client-server` for a reason I think this is ready to merge.

joshix:
@zined Thanks for the contrib!

Issue #832, opened 2016-06-10
checker for coreos.com docs links


kayrus:
Create a tools which checks whether links are valid.

I.e. this doc https://coreos.com/etcd/docs/latest/etcd-live-http-to-https-migration.html has a link to https://coreos.com/etcd/docs/os/generate-self-signed-certificates.html which actually should link to this doc: https://coreos.com/os/docs/latest/generate-self-signed-certificates.html

crawford:
https://crates.io/crates/marker

PR #850, opened 2016-06-29
etcd/live-https-migr: Fix broken links into os/docs/

joshix:
Ref coreos/docs#832


joshix:
Closing with @crawford in re: https://github.com/coreos/docs/pull/798#issuecomment-214518829

backport will go downstream to fix the link/topmatter on prod site, but leaving these links as-is in-repo

PR #852, opened 2016-07-04
Add systemd-timesyncd restart to cloud-config snippet

wkruse:
`systemd-timesyncd` starts before `coreos-cloudinit`. It needs to be restarted in cloud-config in order to apply changed NTP time sources.


zbwright:
Thanks for your work. cloud-config has since been replaced by ignition. closing this PR.

bgilbert:
@wkruse Ignition runs very early in the boot process (from the initramfs) so it shouldn't have this problem.

wkruse:
@bgilbert You're right. Meanwhile we moved to ignition and `systemd-timesyncd` isn't needed anymore.

PR #855, opened 2016-07-06
quay-enterprise/: Update quay container version to current 1.16.3

joshix:
Ref https://github.com/coreos-inc/tectonic-pages/pull/211

Proposed TODO: These should be version "latest", as they are in the
quay-enterprise yaml owned by the tectonic team, see
https://github.com/coreos-inc/tectonic/commit/4f15f8c58321ddf433ceb6ba3838875db2735353#diff-98914fd9e6d06eab312a20550c5fe2df

### DIFF
--- a/quay-enterprise/build-support.md
+++ b/quay-enterprise/build-support.md
@@ -31,7 +31,7 @@ The build worker is currently in beta. To gain access to its repository, please 
 Once given access, pull down the latest copy of the image just like any other:
 
 ```sh
-docker pull quay.io/coreos/registry-build-worker:v1.16.1
+docker pull quay.io/coreos/registry-build-worker:v1.16.3
 ```
 
 ### Run the build worker image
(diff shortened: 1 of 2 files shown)


joshix:
@Quentin-M What was the reasoning behind the change from `latest` to the specific versions in coreos-inc/tectonic-pages#187?

at any rate, "we must update these at release points..." :)

https://github.com/coreos-inc/tectonic-pages/pull/187#issuecomment-224722604

joshix:
https://github.com/coreos-inc/tectonic-pages/issues/186 proposed adding the specific versions

robszumski:
LGTM

Quentin-M:
It was required at least for clair-jwt as the container is built using two different git repositories, both using the `master` branch by default, which is the development branch. The branch had breaking configuration changes and it was also quite confusing to use `latest` (which repository?).

PR #857, opened 2016-07-07
Update ignition examples

alekssaul:
Ignition file creation need to be under "storage" object https://coreos.com/ignition/docs/latest/examples.html


crawford:
Good catch. Can you use two spaces for indentation instead of just one?

alekssaul:
Thanks @crawford, does that work?

joshix:
LGTM

PR #858, opened 2016-07-09
libivirt: updated libvirt deploy script

kayrus:
resolves https://github.com/coreos/bugs/issues/1077


zbwright:
@euank should this still be open?

euank:
While the libvirt script still exists, for the moment we're not linking it anywhere. The documentation does handle the selinux case, so I think it's okay to close this for now and revisit it if/when we bring the script back.

PR #862, opened 2016-07-18
flannel: update docs to use rkt

ajeddeloh:
Update flannel docs to correspond to how flannel is launched in
CoreOS 1097 and newer.

### DIFF
--- a/flannel/flannel-config.md
+++ b/flannel/flannel-config.md
@@ -149,20 +149,13 @@ The last step is to enable `flanneld.service` in the Ignition config:
 
 ## Under the hood
 
-To reduce the CoreOS image size, flannel daemon is stored in CoreOS Enterprise Registry as a Docker container and not shipped in the CoreOS image. For those users wishing not to use flannel, it helps to keep their installation minimal. When `flanneld.service` it started, it pulls the Docker image from the registry. There is, however, a chicken and the egg problem as flannel configures Docker bridge and Docker is needed to pull down the image.
-
-In order to work around this, CoreOS is configured to optionally run a second copy of Docker daemon which we call early-docker. Early-docker daemon is started with `--iptables=false` and containers that it executes need to run with host networking. This prevents Docker from starting `docker0` bridge.
+To reduce the CoreOS image size, flannel daemon is stored in CoreOS Enterprise Registry as an ACI and not shipped in the CoreOS image. For those users wishing not to use flannel, it helps to keep their installation minimal. When `flanneld.service` is started, it pulls the flannel ACI from the registry.
 
 Here is the sequence of events that happens when `flanneld.service` is started followed by a service that runs a Docker container (e.g. redis server):
 
-1. `early-docker.service` gets started since it is a dependency of `flanneld.service`.
-2. `early-docker.service` launches a Docker on a separate Unix socket &mdash; `/var/run/early-docker.sock`.
-3. `flanneld.service` executes `DOCKER_HOST=unix:///var/run/early-docker.sock docker run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (actual invocation is slightly more complex).
-4. flanneld starts and writes out `/run/flannel/subnet.env` with the acquired IP subnet information.
-5. `ExecStartPost` in `flanneld.service` converts information in `/run/flannel/subnet.env` into Docker daemon command line args (such as `--bip` and `--mtu`),
-storing them in `/run/flannel_docker_opts.env`
-6. `redis.service` gets started which invokes `docker run ...`, triggering socket activation of `docker.service`.
-7. `docker.service` sources in `/run/flannel_docker_opts.env` which contains env variables with command line options and starts the Docker with them.
-8. `redis.service` runs Docker redis container.
-
-If you would like to learn more about these service files, you can check them out like so: `systemctl cat early-docker.service`.
+1. `flanneld.service` gets started and executes `/usr/bin/rkt run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (the actual invocation is slightly more complex; the full version can be seen [here](https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel/files/flanneld-rkt.service) or by running `systemctl cat flanneld.service`, which also includes any drop in units).
+2. flanneld starts and writes out `/run/flannel/subnet.env` with the acquired IP subnet information.
+3. `ExecStartPost` in `flanneld.service` converts information in `/run/flannel/subnet.env` into Docker daemon command line args (such as `--bip` and `--mtu`), storing them in `/run/flannel_docker_opts.env`.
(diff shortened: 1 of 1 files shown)


flannel/flannel-config.md
```diff
@@ -149,20 +149,15 @@ The last step is to enable `flanneld.service` in the Ignition config:
 
 ## Under the hood
 
-To reduce the CoreOS image size, flannel daemon is stored in CoreOS Enterprise Registry as a Docker container and not shipped in the CoreOS image. For those users wishing not to use flannel, it helps to keep their installation minimal. When `flanneld.service` it started, it pulls the Docker image from the registry. There is, however, a chicken and the egg problem as flannel configures Docker bridge and Docker is needed to pull down the image.
-
-In order to work around this, CoreOS is configured to optionally run a second copy of Docker daemon which we call early-docker. Early-docker daemon is started with `--iptables=false` and containers that it executes need to run with host networking. This prevents Docker from starting `docker0` bridge.
+To reduce the CoreOS image size, flannel daemon is stored in CoreOS Enterprise Registry as an ACI and not shipped in the CoreOS image. For those users wishing not to use flannel, it helps to keep their installation minimal. When `flanneld.service` it started, it pulls the flannel ACI from the registry.
```
steveeJ: s/it started/is started

ajeddeloh: fixed

flannel/flannel-config.md
```diff
@@ -149,20 +149,15 @@ The last step is to enable `flanneld.service` in the Ignition config:
...
-storing them in `/run/flannel_docker_opts.env`
-6. `redis.service` gets started which invokes `docker run ...`, triggering socket activation of `docker.service`.
-7. `docker.service` sources in `/run/flannel_docker_opts.env` which contains env variables with command line options and starts the Docker with them.
-8. `redis.service` runs Docker redis container.
-
-If you would like to learn more about these service files, you can check them out like so: `systemctl cat early-docker.service`.
+1. `flanneld.service` gets started and executes `/usr/bin/rkt run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (actual invocation is slightly more complex).
+2. flanneld starts and writes out `/run/flannel/subnet.env` with the acquired IP subnet information.
+3. `ExecStartPost` in `flanneld.service` converts information in `/run/flannel/subnet.env` into Docker daemon command line args (such as `--bip` and `--mtu`), storing them in `/run/flannel_docker_opts.env`.
+4. `redis.service` gets started which invokes `docker run ...`, triggering socket activation of `docker.service`.
```
steveeJ: I can't find `redis.service` on recent CoreOS versions, is this outdated?

ajeddeloh: Its mentioned previously in the docs, either ignition or cloudinit creates the systemd unit file for it

flannel/flannel-config.md
```diff
@@ -149,20 +149,15 @@ The last step is to enable `flanneld.service` in the Ignition config:
...
-3. `flanneld.service` executes `DOCKER_HOST=unix:///var/run/early-docker.sock docker run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (actual invocation is slightly more complex).
-4. flanneld starts and writes out `/run/flannel/subnet.env` with the acquired IP subnet information.
-5. `ExecStartPost` in `flanneld.service` converts information in `/run/flannel/subnet.env` into Docker daemon command line args (such as `--bip` and `--mtu`),
-storing them in `/run/flannel_docker_opts.env`
-6. `redis.service` gets started which invokes `docker run ...`, triggering socket activation of `docker.service`.
-7. `docker.service` sources in `/run/flannel_docker_opts.env` which contains env variables with command line options and starts the Docker with them.
-8. `redis.service` runs Docker redis container.
-
-If you would like to learn more about these service files, you can check them out like so: `systemctl cat early-docker.service`.
+1. `flanneld.service` gets started and executes `/usr/bin/rkt run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (actual invocation is slightly more complex).
```
robszumski: Can we link to the actual invocation somewhere, for folks that are interested?

ajeddeloh: We could link them to https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel/files/flanneld-rkt.service but they'd get the same thing by running `systemctl cat flanneld.service`. Do you think its worth it to include that link if it's already on their system?

flannel/flannel-config.md
steveeJ: Isn't that somewhat arbitrary? It's definitely not something you see on a vanilla CoreOS installation. I think it's out of scope for this document.

ajeddeloh: It's just using redis as an example docker container to run; since flannel is used to set up container networking it makes sense have the docs cover what happens when a container is run.

steveeJ: I took the boot sequence out of context of the above example, that's what bothered me about it. While I'd prefer to explain a clean boot sequence that shows only the relation between flannel and docker, it's not in scope of this PR. thanks!

flannel/flannel-config.md
steveeJ: Linking to the service file sounds like a good idea for those who want to know more. also mention the `systemctl cat` command and that it will show the effective service, including drop-ins if installed any.

flannel/flannel-config.md
```diff
@@ -149,20 +149,15 @@ The last step is to enable `flanneld.service` in the Ignition config:
...
+1. `flanneld.service` gets started and executes `/usr/bin/rkt run --net=host quay.io/coreos/flannel:$FLANNEL_VER` (the actual invocation is slightly more complex; the full version can be seen [here](https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel/files/flanneld-rkt.service) or by running `systemctl cat flanneld.service`, which also includes any drop in units).
+2. flanneld starts and writes out `/run/flannel/subnet.env` with the acquired IP subnet information.
+3. `ExecStartPost` in `flanneld.service` converts information in `/run/flannel/subnet.env` into Docker daemon command line args (such as `--bip` and `--mtu`), storing them in `/run/flannel_docker_opts.env`.
+4. `redis.service` gets started which invokes `docker run ...`, triggering socket activation of `docker.service`.
+5. `docker.service` sources in `/run/flannel_docker_opts.env` which contains env variables with command line options and starts the Docker with them.
+6. `redis.service` runs Docker redis container.
+
+If you would like to learn more about these service files, you can check them out like so: `systemctl cat flanneld.service`.
```
steveeJ: this duplicates the comment on https://github.com/coreos/docs/pull/862/files#diff-31d18a18d1bf2a94892dd0c3d49b1025R156

ajeddeloh: Good catch. fixed.

steveeJ:
LGTM

PR #864, opened 2016-07-31
Update auths of docker config

redcancode:
Current stable CoreOS works with this configuration. I also changed the
permissions of the config file to reflect what docker itself uses.

Proof

You can see that the config tests are expecting this format. I couldn't
find any better proof. I tested it myself with CoreOS stable (1068.8.0)
and it works.

https://github.com/docker/docker/blob/v1.10.3/cliconfig/config_test.go#L262

### DIFF
--- a/os/registry-authentication.md
+++ b/os/registry-authentication.md
@@ -10,17 +10,16 @@ Here's what an example looks like with credentials for Docker's public index and
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


josephschorr:
LGTM

robszumski:
LGTM, thanks for updating this!

PR #867, opened 2016-08-08
quay-enterprise/upgrade: Document QE upgrade process

joshix:
Add new document from @ErnestG4 and Ivan with steps to upgrade QE.

### DIFF
--- a/upgrade-quay-enterprise.md
+++ b/upgrade-quay-enterprise.md
@@ -1,0 +1,63 @@
+# Quay Enterprise upgrade guide
+
+This document describes how to upgrade one or more Quay Enterprise containers.
+
+## Backup the Quay Enterprise database
+
+The database is the "source of truth" for Quay, and some version upgrades will trigger a schema update and data migration. Such versions are clearly documented in the [Quay Enterprise release notes][qe-releases].
+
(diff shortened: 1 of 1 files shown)


ErnestG4:
"## Backup the Quay Enterprise database
+
+The database is the "source of truth" for Quay, and some updates will trigger database migrations. Always backup the database before upgrading Quay Enterprise. Once the backup completes, use the procedure in this document to stop the running Quay Enterprise container, start the new container, and check the health of the upgraded Quay Enterprise service.

------------------------------------------------------------------

+## Start the updated Quay Enterprise container
+
+```
+docker run --restart=always -p 443:443 -p 80:80 --privileged=true \
+-v /local/path/to/config/directory:/conf/stack \
+-v /local/path/to/storage/directory:/datastorage \
+-d quay.io/coreos/registry:RELEASE_VERSION
+```
+ Rarely, but occasionally, the new container may perform a database schema upgrade and migration. Versions requiring such database migrations will take a longer time to initially start. These versions are clearly documented in the [release notes](https://tectonic.com/quay-enterprise/releases/), which should be referenced before each QE upgrade"

ErnestG4:
LGTM!

PR #874, opened 2016-08-19
os: add pstore documentation

crawford:
Helps address https://github.com/coreos/bugs/issues/477.


os/collecting-crash-logs.md
```diff
@@ -0,0 +1,94 @@
...
+Panic#2 Part1
+
+$ head --lines=1 /sys/fs/pstore/dmesg-erst-6319986351055831044
+Panic#2 Part2
+
+$ head --lines=1 /sys/fs/pstore/dmesg-erst-6319986351055831045
+Panic#2 Part3
+```
+
+It is important to note that the pstore typically has very limited storage space (on the order of kilobytes) and will not overwrite entries when out of space. The files in `/sys/fs/pstore` must be removed to free up space. The typical approach is to move the files from the pstore to a more permanent storage location on boot, but CoreOS will not do this automatically for you.
```
robszumski: It would be great to give an example of how to do this, but not a blocker.

robszumski:
LGTM

os/collecting-crash-logs.md
cvanlabe: Like!

crawford: Darn. You should be able to capture the console output though, which will have panic traces.

PR #875, opened 2016-08-24
Adding documentation about ETCD_SSL_DIR

venezia:
ETCD_SSL_DIR isn't documented in the coreos docs, but probably should be.


robszumski:
LGTM, thank you for writing this up!

PR #878, opened 2016-08-31
SELinux Docs: copy & paste mistake

pizzarabe:
removed the second 
"$ rm /etc/audit/rules.d/99-default.rules" line


robszumski:
LGTM, thanks for spotting that

PR #882, opened 2016-09-13
Fixed cfssl missing signing artifacts in TL;DR section.

pires:


joshix:
LGTM. Thanks for the fix!

PR #884, opened 2016-09-21
os: state that >1024M is needed for iso/pxe

mischief:


philips:
LGTM

crawford:
This is only true if it's a diskless-PXE boot. I would also say "recommended" instead of "required" since we don't actually know the strict minimum.

PR #888, opened 2016-10-05
os: dns: fix systemd's nss lib name

maximeh:
The plugin for NSS built by systemd is called libnss_resolv**e** and not libnss_resolv, so this configuration should not work.

I haven't checked if in older systemd releases it was called 'resolv' though.

### DIFF
--- a/os/configuring-dns.md
+++ b/os/configuring-dns.md
@@ -4,7 +4,7 @@ By default, DNS resolution on CoreOS is handled through `/etc/resolv.conf`, whic
 
 ## Using a local DNS cache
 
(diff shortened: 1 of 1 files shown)


maximeh:
ping ? Don't mean to be rude, but it's a pretty simple patch (maybe it's wrong, but anyway, some feedback would be appreciated).

Thanks for your efforts.

robszumski:
Sorry for the late response, thank you for bumping this.

maximeh:
No problem, I get it, it's hard to maintain a project :).

PR #890, opened 2016-10-10
iSCSI Docs (First Iteration)

ElijahCaine:
Supersedes #792 

In addition to the original docs I added a short contextual sentence about iSCSI.

As noted in the original PR there are a few improvements that can be made still be made such as:

@robszumski: 

> @joshix Brainstorming some steps to polish:
> - Background
>   - add some general background on how iscsi works, just a sentence or two. It's a block device, etc.
> - Credentials section has examples for global and per-mount
>   - manual examples
>   - cloud-configs (enables iscsid)
> - Expand on the "find available targets"
>   - "You may want to manually search first, then add to a cloud-config..."
> - Mounting instructions
>   - doc has a specific iSCSI example (currently is lacking)
>   - full cloud-config
>   - possibly add a kubernetes example? ([upstream](https://github.com/kubernetes/kubernetes/blob/master/examples/iscsi/iscsi.json))

@mjg59:

> Ah yeah things get a little more complicated if you don't want to do discovery on the machine - the configuration for the target needs to be laid down in that case before login will work. We'll need to provide an example for that.

Thoughts and improvements would appreciated.  We're going to try to merge these ASAP and improve on them later just to get something out the door.

### DIFF
--- a/os/iscsi.md
+++ b/os/iscsi.md
@@ -1,0 +1,54 @@
+# iSCSI on CoreOS
+
+CoreOS has integrated support for mounting iSCSI devices. In order to configure it, do the following:
+
+## Configure the CoreOS iSCSI initiator name
+
+iSCSI clients each have a unique initiator name. CoreOS generates a unique
+initiator name on each install and stores it in
+/etc/iscsi/initiatorname.iscsi. This may be replaced if necessary.
+
+## Configure the global iSCSI credentials
+
+If all iSCSI mounts on a CoreOS system use the same credentials, these may
+be configured locally by editing /etc/iscsi/iscsid.conf and setting the
+node.session.auth.username and node.session.auth.password fields. If the
+iSCSI target is configured to support mutual authentication (allowing the
+initiator to verify that it is speaking to the correct client), these should
+be set in node.session.auth.username_in and node.session.auth.password_in.
+
+## Start the iSCSI daemon
+
+systemctl start iscsid
+
+## Discover available iSCSI targets
+
+Run
+
+iscsiadm -m discovery -t sendtargets -p target_ip:port
+
+## To provide target-specific credentials
+
+You can replace the global credentials with target-specific credentials by running:
+
+iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.username --value=username
+
+iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.password --value=password
+
+## Log into an iSCSI target
+
+iscsiadm -m node --login
+
+will log into all discovered targets. To log into a specific target:
+
+iscsiadm -m node --targetname=targetname --login
+
+## Enable automatic iSCSI login at boot
+
+If you want to connect to iSCSI targets automatically at boot, run
+
+systemctl enable iscsid
+
+## Mounting iSCSI targets
+
+See https://coreos.com/os/docs/latest/mounting-storage.html
--- a/os/iscsi.md
+++ b/os/iscsi.md
@@ -88,6 +88,7 @@ To create and enable the iscsid service automatically using cloud-config and ign
 <!-- TODO: It's inclear based on documentation what the actual first line of this doc snippet should be.
      This is a best guess based on docs I've read, the rest I'm pretty certain of.
      I know we want to do discovery in this file, just not sure if that line accomplished the task. -->
+     
 ```
 isns.address = host_ip
 isns.port = host_port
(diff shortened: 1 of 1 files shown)


os/iscsi.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Start the iSCSI daemon
+
+```
+systemctl start iscsid
+```
+
+## Discover available iSCSI targets
+
+Run
```
robszumski: Single word looks out of place, what about "To discover targets, run:"

os/iscsi.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+iscsiadm -m discovery -t sendtargets -p target_ip:port
```
robszumski: Add dollar sign and `sh` highlighting hint

```sh
$ iscsiadm -m discovery -t sendtargets -p target_ip:port
```

os/iscsi.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## To provide target-specific credentials
+
+You can replace the global credentials with target-specific credentials by running:
+
+```
+iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.username --value=username
```
robszumski: dollar signs

robszumski: (Add these everywhere)

robszumski: Do you fill in a name for `--targetname=targetname`? Can we use a real example or somehow indicate that this shouldn't be the same value as the flag name.

os/iscsi.md
```diff
@@ -0,0 +1,64 @@
...
+iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.password --value=password
+```
+
+## Log into an iSCSI target
+
+```
+iscsiadm -m node --login
+```
+
+will log into all discovered targets. To log into a specific target:
```
robszumski: Can we move "will log into all discovered targets." above this, so it reads better?

os/iscsi.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+iscsiadm -m node --targetname=targetname --login
+```
+
+## Enable automatic iSCSI login at boot
+
+If you want to connect to iSCSI targets automatically at boot, run
```
robszumski: Maybe give a cloud-config or ignition example for this?

ElijahCaine: I agree that this section should be expanded a bit more, but I'm not sure the best way to generate/obtain a cloud-config / ignition file example.  Any ideas where I should start looking / asking for one?

os/iscsi.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+$ iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.username --value=username
+$ iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.password --value=password
+```
+
+`targetname`, `username`, and `password` are specific to you, not literal strings. For example:
+
+```
+$ iscsiadm -m node --targetname=custom_target --op-update --name=node.session.auth.username --value=some_username
```
robszumski: Why have both of these examples, when one is real and one is more fake. Should we just remove the fake one?

Both reading the long list of flags, as well as spotting the differences between the two are kinda hard. What about this:

## Provide target-specific credentials

For each unique `--targetname`, first enter the username:

```
$ iscsiadm -m node \
  --targetname=custom_target \
  --op-update \
  --name=node.session.auth.username \
  --value=myuser
```

And then the password:

```
$ iscsiadm -m node \
  --targetname=custom_target \
  --op-update \
  --name=node.session.auth.password \
  --value=p@ssw0rd
```

os/iscsi.md
robszumski: Reading the docs is your best bet. But the main strategy is probably to write out a config file (`/etc/iscsi/iscsid.conf`) that are equivalent to the above commands. Then start the `iscsid` unit via cloud-config and ignition.

https://coreos.com/os/docs/latest/cloud-config.html
https://coreos.com/ignition/docs/latest/

I can help you out once you check out the docs. Let me know!

robszumski: Layout wise, you can see how the mounting storage docs include inline cloud-config examples: https://coreos.com/os/docs/latest/mounting-storage.html

os/iscsi.md
```diff
@@ -0,0 +1,80 @@
...
+```
+$ iscsiadm -m discovery -t sendtargets -p target_ip:port
+```
+
+## To provide target-specific credentials
+
+You can replace the global credentials with target-specific credentials by running:
+
+```
+$ iscsiadm -m node --targetname=targetname --op-update --name=node.session.auth.username --value=username
```
mjg59: Should be --op update, not --op-update - same for the next few lines

os/iscsi.md
ElijahCaine: @robszumski I did my best to figure out what a cloud-config, ignition, and iscsid.conf would look like.  I don't have an easy way to test these docs right now.  Let me know if the changes I just pushed aren't right.

robszumski:
Added two commits:

1. Wrote out the config file using cloud-config/ignition. This is the only way these would work fully automatically at machine provisioning time. I included the most minimal config that should work, which was similar to what you had. Note that I added `#cloud-config`, which is the only way that config is valid. You can always check with https://coreos.com/validate. Your ignition config had an extra trailing comma, which was also invalid.
2. Re-ordered the hierarchy to put manual vs automatic as the top level choice. This matches some of the other docs, with the manual steps first (as you had) and the automatic version of the same at the end. Under auto, the choice is then cloud-config or ignition. 

Let me know what you think of those changes.

robszumski:
I didn't have time to test any of this, but a good way would be to use Vagrant with two machines. One serving iSCSI out of a docker container backed by local storage, and another provisioned to test both of the automatic configs, attempting to mount the target from the first machine.

ElijahCaine:
@robszumski Thank you for the commit.  I'll bookmark that link for validating config files.  I'll also try to test this doc asap to make sure it's correct.

ElijahCaine:
@robszumski As a heads up, I'm doing my best but I really don't know the best way to test this document.  I've gotten to the point where I have two CoreOS vagrant boxes hooked up with the cloud-config examples in the doc, but I have no idea what to do next to *actually* prove document correctness.

I don't really want to publish a doc that's not 100% accurate, but I also thing *any* doc is better than *no* doc.  If a *different* intern (or person with time to spare and adequate knowledge to do get the job done) would like to test this that'd be cool, but for the time being I really can't guarantee I'll be able to test and make corrections in a timely manner.

robszumski:
I guess let's just get this merged...objections?

PR #414, opened 2015-01-09
config dir and SELinux

ibukanov:
With SElinux in enforcing mode it is necessary to label manually the config dir on the host so CoreOS can access it.

PR #891, opened 2016-10-10
Rebased config dir and SELinux

ElijahCaine:
This is a rebased version of #414.

@ibukanov:
> With SElinux in enforcing mode it is necessary to label manually the config dir on the host so CoreOS can access it.

#414 was given a +1 by @marineam.  Let's merge it this time!


robszumski:
LGTM

PR #892, opened 2016-10-11
os/network-config-with-networkd: add supported ways to disable IPv6

dm0-:
Document the systemd-approved way to disable IPv6 since the upstream fix for coreos/bugs#1419 was closed.


robszumski:
LGTM!

PR #896, opened 2016-10-13
Fix email in key generation script

ErnestG4:
The email used in generating a public signing key should be pulled from aci-signing-key-batch.txt

Changed to pull from the second entry of the row with a first entry of "Name-Email:" in aci-signing-key-batch.txt.


robszumski:
LGTM

PR #898, opened 2016-10-18
os: add section on visudo usage to adding-users doc

kbrwn:
Visudo on CoreOS requires the file be specified or an error is thrown.
Issue: https://github.com/coreos/bugs/issues/1617

### DIFF
--- a/os/adding-users.md
+++ b/os/adding-users.md
@@ -66,7 +66,7 @@ update-ssh-keys -u user1 user1.pem
 
 ## Grant sudo Access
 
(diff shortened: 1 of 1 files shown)


os/adding-users.md
```diff
@@ -64,6 +64,33 @@ To assign an SSH key, run:
 update-ssh-keys -u user1 user1.pem
 ```
 
+## Grant sudo Access
+
+If you trust the user you can grant administrative privileges using visudo. visudo checks the file syntax before actually overwriting the sudoers file. This command should be ran as root to avoid losing sudo access in the event of a failure. Instead of editing /etc/sudo.conf directly you will create a new file under the `/etc/sudoers.d/` directory. When you run visudo it is required that you specify which file you are attempting to edit with the -f argument:
```
joshix: backticks: `visudo`.
backticks: `sudoers`

crawford: "If you trust the user, you..."

crawford: "This command should be run as root..."

crawford: "When you run visudo, it..."

crawford:
Can you please rebase before merging?

PR #905, opened 2016-10-26
qe: revamp section on manually adding license file

kbrwn:
Emphasis on using the management panel over the manual option by listing it first in the doc.
Adds an example of what the config dir needs to look like before the QE container is started.


robszumski:
LGTM. I added a commit that tweaked a formatting issue.

PR #906, opened 2016-10-26
quay-enterprise: fix markdown formatting

robszumski:


robszumski:
Quick change to fix some formatting

PR #912, opened 2016-11-03
quay-enterprise: remove bittorrent.md

kbrwn:
This doc is confusing for end-users as QE support for BitTorrent has not been shipped. 

https://github.com/coreos-inc/quay/pull/1872


kbrwn:
hmm didn't mean to include the first two commits...

PR #913, opened 2016-11-04
Google Analytics Beacon Tracking Pixel & Script

ElijahCaine:
Includes:
- Script that ensures the existance of the footer.
- Footer added to all `*.md` files

Footer is not correct yet:
- Mising ga-beacon instance url
- Missing Google Analytics tracking ID

CC @joshix @robszumski @philips


ElijahCaine:
@joshix @crawford I'd love to know what you think we should do about this. I know you don't want to dirty up the docs with scripts but we need a script here to ensure we have the analytics tracking pixel at the bottom of all of the docs.
We will also need to add *something* to our build system that strips that pixel out of the docs before building.

ElijahCaine:
This is on ice for a while until we get the tracking pixel merged into another upstream project's documentation.

zbwright:
@robszumski should this PR still be open? or has this been resolved?

PR #921, opened 2016-11-16
Docs: Clair trust to Self-signed QE

alekssaul:
Outline how clair-jwt container can trust self-signed QE registry.

/cc: @josephschorr , @eddieesquivel

### DIFF
--- a/quay-enterprise/clair.md
+++ b/quay-enterprise/clair.md
@@ -218,7 +218,7 @@ If your certificates use a public CA, you are now ready to run Clair. If you are
 Similar to the process for setting up Docker to [trust your self-signed certificates][registry-self-signed], Clair must also be configured to trust your certificates. Using the same CA certificate bundle used to configure Docker, complete the following steps:
 
 1. Rename the same CA certificate bundle used to set up Quay Registry to `ca.crt`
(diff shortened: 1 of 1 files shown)


quay-enterprise/clair.md
```diff
@@ -212,6 +212,16 @@ To configure Clair to run under TLS, a few additional steps are required:
 2. Place these files as `clair.crt` and `clair.key` in your Clair configuration directory
 3. Uncomment the `key_file` and `crt_file` lines under `verifier_proxies` in your Clair `config.yaml`
 
+### Configuring trust to self-signed SSL
+
+To configure Clair with a Quay Registry that is configured with self-signed SSL certificates follow steps below:
+
+1. Rename the Quay Registry trusted certificate bundle to ca.crt
```
robszumski: Where do you get this file from in the first place?

alekssaul: Customer needs to bring the file however they sign the Certificate. If they are self-signing with something like openssl, they would follow something like; https://tectonic.com/quay-enterprise/docs/latest/quay-ssl.html . If they use something like Active Directory Certificate Authority, they would follow Microsoft Documentation to do so.

PR #928, opened 2016-11-21
os: add device size requirements

robszumski:
User on IRC:

> alright yeah, that was it. used a bigger card and no problems. maybe "you'll get this cryptic error message if the device is too small" should be added to those docs? I feel like installing a server OS on 4 GB probably isn't that uncommon.

### DIFF
--- a/os/installing-to-disk.md
+++ b/os/installing-to-disk.md
@@ -2,7 +2,7 @@
 
 ## Install script
 
-There is a simple installer that will destroy everything on the given target disk and install CoreOS. Essentially it downloads an image, verifies it with gpg, and then copies it bit for bit to disk.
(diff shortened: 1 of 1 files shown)


fletom:
sounds good!

if you want, you can also include the cryptic error message you get when your drive isn't big enough:

```Error: return code 2 from ROOT_DEV=$(blkid -t "LABEL=ROOT" -o device "${DEVICE}"*)```

either way, I'm putting it here for posterity

File: os/installing-to-disk.md [hunk 1]
```diff
@@ -2,7 +2,7 @@
 
 ## Install script
 
-There is a simple installer that will destroy everything on the given target disk and install CoreOS. Essentially it downloads an image, verifies it with gpg, and then copies it bit for bit to disk.
+There is a simple installer that will destroy everything on the given target disk and install CoreOS. Essentially it downloads an image, verifies it with gpg, and then copies it bit for bit to disk. An installation requires at least 4.5 GB of usable space on the device, although CoreOS doesn't currently utilize all of it.
```
crawford: It does use all of it though. I thought we specified that the minimum disk disk was 8GB somewhere else. That is the number I would actually like to go with.

PR #930, opened 2016-11-26
*: fix broken links

crawford:
Courtesy of the new link validation program.

PR #933, opened 2016-11-30
available in japanese 404 fix

bhaggerty:
Fixing Link for "available in Japanese"


crawford:
Also fixed in https://github.com/coreos/docs/pull/930. You're killing me, smalls.

PR #938, opened 2016-12-08
Quay Enterprise - pull secret documentation

eddieesquivel:
Tectonic.com no longer offer dockercfg files. Only offers config.json. Updated the documentation to reflect this change.


quay-enterprise/initial-setup.md
```diff
@@ -31,7 +31,22 @@ sudo docker run -d -p 6379:6379 quay.io/quay/redis
 
 ## Downloading the Quay Enterprise image
 
-After signing up you will receive a `.dockercfg` file containing your credentials to the `quay.io/coreos/quay` repository. Save this file to your CoreOS machine in `/home/core/.dockercfg` and `/root/.dockercfg`. You should now be able to execute `docker pull quay.io/coreos/quay:v2.0.1` to download the container.
+After signing up you will be able to download a pull secret that you can rename to `config.json`.
```
joshix: just one question for clarity -- isn't `config.json` still the dockercfg format, so you should select that when downloading? (Cf. https://github.com/coreos/docs/blob/master/quay-enterprise/tectonic/index.md)

alekssaul: When I think of dockercfg it's the "old file format" (I believe pre Docker v1.07) that gets placed into ~/.dockercfg as opposed to new one in "~/.docker/config.json" . Config.json supports HttpHeaders property and auth property. 

I believe at some point tectonic.com served them both.

PR #945, opened 2016-12-08
fleet functionality K8s does not have

ElijahCaine:
This is a work in progress. Feedback and discussion encouraged.

References documentation needed to transition toward Kubernetes from fleet.

cc: @joshix


mischief:
s/depricated/deprecated/

kubernetes/depricated-functionality.md
```diff
@@ -0,0 +1,25 @@
+# Depricated functionality
+
+Despite their many functional similarities, some features just don't exist in in Kubernetes like they did in the fleet.
```
lucab: "in in"

ElijahCaine:
So the last header of this doc "ExecStop" kinda does exist via the pod [spec.container.lifecycle.exec](http://kubernetes.io/docs/user-guide/pods/#termination-of-pods). I'll mention that there's no explicit way to tell Kubernetes how to shutdown your container but you can pre-empt it with that.

philips:
This content seems fine.

kubernetes/deprecated-functionality.md
```diff
@@ -0,0 +1,65 @@
+# Deprecated functionality
+
+While fleet and Kubernetes are largely similar in functionality, some specific fleet features do not have direct equivalents in Kubernetes.Fortunately, workarounds exist for almost every use-case. Several of these features and workarounds are outlined below.
+
+## Container Dependencies
+
+Fleet uses systemd service dependencies to outline a *limited* dependency graph. When units are co-located then containers may be specified to start in a specific order; only beginning a service *after* others it depends on have begun. This is not really a fleet feature but rather a systemd feature; it is limited by the design and feature-set of systemd
+
+While this has been [discussed at length in the Kubernetes community][pod-deps-discussion] it has not been implemented in Kubernetes as of early 2017. There are two workarounds for this in Kubernetes:
```
philips: I think we should essentially say most use cases can be solved by crash loops on a pod and init containers. I don't think pod deps will ever be implemented.

kubernetes/deprecated-functionality.md
```diff
@@ -0,0 +1,65 @@
...
+        exec:
+          # SIGTERM triggers a quick exit; gracefully terminate instead
+          command: ["/usr/sbin/nginx","-s","quit"]
+```
+
+[comment]: # (TODO: Think of a better example for preStop. Need a clearer demo.)
+
+More information can be found at the [Kubernetes Pods user guide][prestop]
+
+**Note:** The `preStop` directive is not a replacement for ExecStop. After a grace-period the container will still be killed via `SIGTERM`.
```
philips: this isn't totally true. You can set the graceful shutdown time to nearly forever terminationGracePeriodSeconds

https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_pod

joshix:
superseded in https://github.com/coreos/docs/pull/991

PR #949, opened 2016-12-15
quay-enterprise: update initial-setup.md to most recent release

kbrwn:
Updates setup document to use latest quay enterprise release. `v2.0.3` includes several bug fixes. We've received a high number of tickets that are due to this doc being tagged at `v2.0.1`


robszumski:
Ideally all of these are updated during the release process

PR #950, opened 2016-12-16
dead link

stephanlindauer:
relative path won't work here.
currently https://coreos.com/kubernetes/docs/latest/network-troubleshooting.html redirects to 404 on https://coreos.com/os/docs/latest/docs/latest/install-debugging-tools.html
no sure if this will fix it though...


crawford:
Links in the source need to work properly. This PR breaks that. The site uses a sync script to rewrite the links and it looks like we just need to run the more recent version against the existing docs.

Thanks anyway.

PR #951, opened 2016-12-21
os/sdk: instructions for building an AMI

euank:
This is what I did to get a local dev image over to the Great Amazon Cloud:tm:. I couldn't find any existing instructions, so here we are.

If any of the steps are missing something or not up to date with what we actually do, I'd be happy to improve them.

cc @crawford @marineam since I think you have most of the knowledge about what a sane dev workflow for ami-building is.

### DIFF
--- a/os/sdk-building-aws-images.md
+++ b/os/sdk-building-aws-images.md
@@ -1,0 +1,69 @@
+# Building AWS images
+
+## Build the image
+
+Prior to creating an AMI, you must have built the image in the first place.
+
+Follow the instruction for building [production images][prodimages] or [development images][devimages].
+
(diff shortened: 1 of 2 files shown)


marineam:
There is a newer script, `import.sh`, which is what we currently use. Instead of directly writing to an EBS volume it uploads to S3 and uses a newer API that can import disks from S3 to EBS. So an EC2 instance isn't required but probably still recommended since the upload to S3 is an uncompressed disk image. Otherwise the procedure is similar.

euank:
Awesome, I'll go through that flow and update this PR based on that. Thanks!

euank:
Closing, this should all be replaced with ore and might as well happen in a new pr.

PR #956, opened 2016-12-28
*: rename coreos to container linux

crawford:


robszumski:
LGTM. Not a single instance of CoreOS was replaced in the wrong context, I'm impressed.

eg. The engineers at Container Linux have worked hard...

crawford:
Yeah, this was a pain. I actually had to read every occurrence. I'm impressed that you read all of them as well. Let's not rename the OS anymore.

PR #957, opened 2016-12-29
os/rollback: update instructions for dm-verity

euank:
The instructions made assumptions that are not true when running with
dm-verity.

This updates the instructions to something much more verbose, but that
works for systems with or without dm-verity

See https://github.com/coreos/bugs/issues/1718

@crawford

### DIFF
--- a/os/manual-rollbacks.md
+++ b/os/manual-rollbacks.md
@@ -5,7 +5,7 @@ As much as we all love flawless, automatic updates, there may be occasions when 
 tl;dr: The following command will set the currently passive partition to be active on the next boot:
 
 ```
-cgpt prioritize "$(cgpt find -t coreos-usr | grep --invert-match "$(findmnt --noheadings --raw --output=source --target=/usr)")"
+cgpt prioritize "$(cgpt find -t coreos-usr | grep --invert-match "$(rootdev -s /usr)")"
 ```
 
 ## How do updates work?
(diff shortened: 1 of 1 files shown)


euank:
I've manually verified that the "cgpt find ..." bit works as I expect on both 1192.0.0 (no dm-verity) and 1262.0.0 (with dm-verity).

These are the assumptions I know are being made:
1. Only one active usr partition has a label (e.g. `/dev/vda3` is "USR-A" and `/dev/mapper/usr` has no partlabel)
2. dm-verity does not obscure the uuid of the partition (e.g. `lsblk /dev/vda3` and `/dev/mapper/usr` have the same uuid)

In practice, both of those seem to be true.

The second assumption would break if we added lvm into the mix pretty much anywhere.

os/manual-rollbacks.md
```diff
@@ -110,17 +110,25 @@ $ cgpt find -t coreos-usr
...
+$ lsblk --output NAME,TYPE,MOUNTPOINT
+NAME    TYPE  MOUNTPOINT
+sda     disk  
+...
+|-sda4  part  
+| `-usr crypt /usr
+...
 ```
 
+Note: Older Container Linux systems will not have an additional "crypt" partition.
```
robszumski: Can we explain why or what it is? Any maybe attempt to rephrase so it doesn't have to say `Note:`?

marineam:
Take a look at the rootdev tool, it can resolve the backing device for any given mount point.

euank:
@marineam unfortunately it doesn't get what I'm looking for:

```shell
$ findmnt --noheadings --raw --output=source /usr
/dev/mapper/usr
$ rootdev /usr
/dev/dm-0
$ echo /dev/$(lsblk --list --output-all --json | jq -r '.blockdevices as $bd | $bd[] | select(.mountpoint == "/usr") | .uuid as $uuid | $bd[] | select(.uuid == $uuid).kname | select(startswith("dm") != true)')
/dev/vda3
```

If there's a better way to go from `/usr` -> `/dev/vda3` in the face of dm-verity, I'd love to know it since that mess of lsblk + jq looks way too complicated, but I don't think rootdev can do it.

euank:
Thanks @marineam! Updated per your suggestion

os/manual-rollbacks.md
```diff
@@ -5,7 +5,7 @@ As much as we all love flawless, automatic updates, there may be occasions when
 tl;dr: The following command will set the currently passive partition to be active on the next boot:
 
 ```
-cgpt prioritize "$(cgpt find -t coreos-usr | grep --invert-match "$(findmnt --noheadings --raw --output=source --target=/usr)")"
+cgpt prioritize "$(cgpt find -t coreos-usr | grep --invert-match "$(rootdev /usr)")"
```
marineam: `rootdev -s /usr`

euank:
Whoops, thanks for catching that, fixed

PR #962, opened 2017-01-04
*: fix broken headings

euank:
As best I can tell, these are all the headings in this repo that were
missing a newline.


crawford:
@euank mind adding a rule to marker?

robszumski:
I'm open to fixing this in Jekyll, but afaik we are using the exact same engine that Github pages uses, and hence works with the previews on Github.

euank:
@crawford I think by common-mark it's fine either way, so it doesn't really make sense for marker to care.

If we did, it would presumably be a warning that it might not behave well for other markdown flavors, but this does seem more like a jekyll issue than a marker thing.

crawford:
@euank According to the [dingus](http://spec.commonmark.org/dingus/?text=%60%60%60%0A%5Bmdlinks%5D%3A%20http%3A%2F%2Fdaringfireball.net%2Fprojects%2Fmarkdown%2Fsyntax%23link%20%22Markdown%20link%20syntax%22%0A%5Breadme%5D%3A%20README.md%0A%60%60%60%0A%23%23%23%23%20Relative%20URLs%20preferred), it is valid. Nevermind!

PR #963, opened 2017-01-04
os: update azure image name

crawford:
This was re-published with the new name.


mischief:
@crawford mantle should likely be updated for this too if it hasn't already.

crawford:
I'm still testing those changes :)

crawford:
Verified that this worked.

PR #971, opened 2017-01-13
Removed ellipses to improve formatting

wsanders:
os/booting-with-pxe.md line 186: I removed the ellipses and newline so the https://coreos.com/os/docs/latest/booting-with-pxe.html generated from this would render the initrd example textbox properly. With the ellipses (or maybe it was the newline) the kernel and initrd lines were joined together into a single line in my browser.

### DIFF
--- a/os/booting-with-pxe.md
+++ b/os/booting-with-pxe.md
@@ -178,11 +178,13 @@ usr/share/oem
 usr/share/oem/example.ign
 ```
 
-Add the `oem.cpio.gz` to your PXE boot directory and [append it][append-initrd] to the `initrd` line in your `pxelinux.cfg` and specify the path to the config:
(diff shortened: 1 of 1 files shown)


os/booting-with-pxe.md
```diff
@@ -179,11 +179,10 @@ usr/share/oem/example.ign
 ```
 
 Add the `oem.cpio.gz` to your PXE boot directory and [append it][append-initrd] to the `initrd` line in your `pxelinux.cfg` and specify the path to the config:
+
```
robszumski: the previous lack of new line is was causing the formatting error. We can choose to keep the ellipses or not, what do you prefer?

wsanders:
No preference...

os/booting-with-pxe.md
joshix: Let's keep the ellipses and add the newline

joshix:
@wsanders Note that this will be fixed on coreos.com/docs after the next release. Thanks for the fix.

PR #981, opened 2017-01-27
etcd: remove language about redirects

robszumski:
Fixes issue on etcd-dev, where etcd 2.x no longer requires curl to follow redirects, since these are handled internally by etcd.

cc: @joshix @ElijahCaine


ElijahCaine:
LGTM

PR #994, opened 2017-02-13
Update booting on interoute

phillipkent:
This is an update for booting on Interoute VDC in its current configuration.


robszumski:
Since I see multiple commits, let me know when you are done and we can merge.

phillipkent:
Hi, it is ready to merge.

PR #997, opened 2017-02-15
os/booting-on-ecs.md: systemd and contents are not valid

anvaughan:
`systemd` and `contents` are not valid on CoreOS-stable-1235


crawford:
This config cannot be passed directly to the userdata. It needs to be transformed by ct first. Were you reading this doc on https://coreos.com or on GitHub?

PR #1003, opened 2017-02-22
q-e/: Update qe container version 2.1.0

joshix:


joshix:
(forgetting to do this upstream is why the downstream version always looks so drastic)

PR #1008, opened 2017-03-02
os:fix customizing docker cloud-config

flybyray:
permissions value under write_files should not be treated as numbers.
Fixes #1839. https://github.com/coreos/bugs/issues/1839


crawford:
The existing documentation is fine (just tested it). Why is this change needed?

PR #1015, opened 2017-03-08
STYLE: Update GIFEE expansion. Add capitalization rules for project names

joshix:
Proposed. List of projects with lower case rule can probably be lengthened.

Eventually want to simplify this rule to be "Project names are proper nouns." Currently proposing we write them with initial capitals when beginning a sentence. Baby steps.


joshix:
@jonboulle I know you are going to be this PR's biggest fan. PTAL.

Issue #989, opened 2017-02-06
CoreOS Container Linux etcd setup guide missing?


urzds:
I am looking at https://coreos.com/etcd/docs/latest/dl_build.html. In the left navigation panel I cannot find a link to a guide how to setup etcd on CoreOS Container Linux. There are all kinds of guides for running it manually, but there seems to be none explaining e.g. the `etcd-member.service` systemd.unit.

philips:
Yes, we need to fix this ASAP. cc @joshix @coreos/team-os 

The docs are at the bottom of this blog post: https://coreos.com/blog/toward-etcd-v3-in-container-linux.html

```
$ systemctl enable /usr/lib/coreos/etcd-member.service
$ systemctl start etcd-member
$ rkt list
UUID		APP	IMAGE NAME			STATE	CREATED		STARTED		NETWORKS
4d566696	etcd	quay.io/coreos/etcd:v3.0.10	running	6 minutes ago	6 minutes ago
$ etcdctl set fname Regis
$ etcdctl get fname
Regis
```

It needs to be cross linked from the following locations:

- https://coreos.com/etcd/docs/latest/getting-started-with-etcd.html

And needs to replace this doc:

- https://coreos.com/etcd/docs/latest/v2/docker_guide.html

joshix:
fixed (a bit ago) in https://github.com/coreos/docs/commit/37312455f993a8595277aa39d440fabfa2a791d3

expanded in https://github.com/coreos/docs/pull/1016

_published_ in 3.1.3: https://coreos.com/etcd/docs/latest/getting-started-with-etcd.html

joshix:
closed fixed (MVP)

PR #1016, opened 2017-03-08
etcd: 'etcd-member.service' in Container Linux

gyuho:
Fix https://github.com/coreos/etcd/issues/7436.

Also address https://github.com/coreos/docs/issues/989.

/cc @euank @crawford @xiang90

### DIFF
--- a/etcd/getting-started-with-etcd.md
+++ b/etcd/getting-started-with-etcd.md
@@ -44,6 +44,50 @@ etcd v3 startup can be configured at a new node's first boot with [Ignition][ign
 }
 ```
 
+`etcd-member.service` has some default settings
+
+```bash
+$ systemctl cat etcd-member.service
+
+<<COMMENT
+...
+Environment="ETCD_IMAGE_TAG=v3.0.10"
+Environment="ETCD_NAME=%m"
+Environment="ETCD_USER=etcd"
+...
+COMMENT
+```
+
+Default configurations can be overwritten with `1-override.conf` file, as below:
+
+```bash
+# to update etcd-member service file
+$ cat > /tmp/override-my-etcd.conf <<EOF
+[Service]
+Environment="ETCD_IMAGE_TAG=v3.1.2"
+Environment="ETCD_DATA_DIR=/var/lib/etcd"
+Environment="ETCD_SSL_DIR=/etc/ssl/certs"
+Environment="ETCD_OPTS=--name my-etcd-1 \
+    --listen-client-urls https://10.240.0.1:2379 \
+    --advertise-client-urls https://10.240.0.1:2379 \
+    --listen-peer-urls https://10.240.0.1:2380 \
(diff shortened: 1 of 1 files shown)


etcd/getting-started-with-etcd.md
```diff
@@ -44,6 +44,50 @@ etcd v3 startup can be configured at a new node's first boot with [Ignition][ign
...
+    --advertise-client-urls https://10.240.0.1:2379 \
+    --listen-peer-urls https://10.240.0.1:2380 \
+    --initial-advertise-peer-urls https://10.240.0.1:2380 \
+    --initial-cluster my-etcd-1=https://10.240.0.1:2380,my-etcd-2=https://10.240.0.2:2380,my-etcd-3=https://10.240.0.3:2380 \
+    --initial-cluster-token my-etcd-token \
+    --initial-cluster-state new \
+    ...
+EOF
+$ mkdir -p /etc/systemd/system/etcd-member.service.d
+$ mv /tmp/override-my-etcd.conf /etc/systemd/system/etcd-member.service.d/override.conf
```
lucab: I think we usually name those `nnn-label.conf` to make clear multiple dropins are loaded in lexicographic order. I.e. `10-override.conf` here.

etcd/getting-started-with-etcd.md
```diff
@@ -44,6 +44,50 @@ etcd v3 startup can be configured at a new node's first boot with [Ignition][ign
...
+    --initial-cluster my-etcd-1=https://10.240.0.1:2380,my-etcd-2=https://10.240.0.2:2380,my-etcd-3=https://10.240.0.3:2380 \
+    --initial-cluster-token my-etcd-token \
+    --initial-cluster-state new \
+    ...
+EOF
+$ mkdir -p /etc/systemd/system/etcd-member.service.d
+$ mv /tmp/override-my-etcd.conf /etc/systemd/system/etcd-member.service.d/1-override.conf
+
+# to check service-file-override status
+$ systemd-delta --type=extended
```
euank: I think `systemctl cat etcd-member` is more readable, but *shrug

etcd/getting-started-with-etcd.md
```diff
@@ -44,6 +44,50 @@ etcd v3 startup can be configured at a new node's first boot with [Ignition][ign
...
+Environment="ETCD_USER=etcd"
+...
+COMMENT
+```
+
+Default configurations can be overwritten with `1-override.conf` file, as below:
+
+```bash
+# to update etcd-member service file
+$ cat > /tmp/override-my-etcd.conf <<EOF
```
euank: I've always wondered why we do this cat/mkdir/mv song and dance in our docs vs `systemctl edit etcd-member`.

Is there a particular reason here?

gyuho: No particular reason, I just find it easy to move around those when `sudo` is needed.

PR #1026, opened 2017-03-23
os/manual-rollbacks: Add note that arbitrary version selection is not supported

zbwright:


os/manual-rollbacks.md
```diff
@@ -2,6 +2,8 @@
 
 As much as we all love flawless, automatic updates, there may be occasions when an update must be rolled back. This is fairly straightforward on Container Linux, once you know the right commands.
 
+**Note**: Container Linux allows you to rollback to a previously active partition, or change to an older release channel (to downgrade to a stable version). It does not allow you to downgrade to an uninstalled, user defined version number.
```
joshix: This doesn't quite capture it. I think we're after something like:

"The active version of Container Linux can be rolled back to the version installed on the rollback partition immediately after an upgrade, or downgraded to the version current on any lower release channel. Rolling back to an arbitrary previous version number is not supported."

joshix: I also half-suspect (not sure, suggesting here) that the location for this note might be _after_ the two rollback procedures, as a final clarifier.

zbwright: I think the note goes right at the top. People hitting this page should know that there are limited rollback options. And I think the note was added specifically because users were calling tech support, complaining that rolling back to a defined version number wasn't working as expected.

question - can be rolled back to a version installed immediately after an upgrade? or can be rolled back, immediately after an upgrade, to the version installed on the rollback partition?

The active version of Container Linux can be rolled back to the version installed on the rollback partition immediately after an upgrade, or downgraded to the version current on any lower release channel. Rolling back to an arbitrary previous version number is not supported.

joshix: Your last graf is almost perfect, but: you can roll back to the rollback partition _after_ an upgrade reboot, rather than before one.

zbwright: Immediately after an upgrade reboot, the active version of Container Linux can be rolled back to the version installed on the rollback partition, or downgraded to the version current on any lower release channel. Rolling back to an arbitrary previous version number is not supported.

euank: One possible issue is that saying rolling back to an arbitrary version is not supported sorta implies that the means of rolling back mentioned here are. They're not.

The only supported rollback is the "new partition won't boot, rollback before it boots" case (which happens automatically). Any time a version of Container Linux successfully boots, rolling back to a previous version (whether by channel downgrade or switching usr partitions), is no longer supported.

zbwright: If that's true, this whole page should be repositioned - and it's not just a matter of adding a note at the top. As it is, the page doesn't describe a 'boot failed, you must rollback' process. It seems to describe a rollback after a successful update. Please confirm, and I'll take a stab at it.

I will ask 1) whether the rollback happens automatically, and 2) if not, what's the error message presented when reboot fails?

joshix: @euank Good and fine distinction. 

Last sentence: "There is no method for rolling back to an arbitrary previous version number."

joshix: @zbwright Just to recite the cases in case we've caused confusion:

Suppose: A new version is released on the Alpha channel and my Container Linux node updates.

1. If the update fails, automatic rollback to B partition (supported; automatic)
OR
2. Update succeeds and I run on with the new version number (supported)
OR
3. Update succeeds, but I roll back to whatever version is on my B partition as described in this doc (you're on your own)
OR
4. Update succeeds, but I choose a lower channel and "update down" to whatever version number is current on that channel (unsupported)

zbwright:
this PR has been moved to  #1034 . (I broke this one, and created 1034 to capture progress.)

joshix:
closed superseded in https://github.com/coreos/docs/pull/1034

PR #999, opened 2017-02-15
os: add missing example of file contents

robszumski:
Fixes https://github.com/coreos/docs/issues/998

PR #1035, opened 2017-04-07
os/env: remove spurious ExecStart

lucab:


lucab:
/cc @crawford @robszumski 

I guess the empty property is a spurious line introduced in #999?

crawford:
This is needed to clear the previous ExecStart, no?

lucab:
I see. But then I'm confused about systemd.service doc which seems to allow this only for `Type=oneshot`.

crawford:
Because `Type=oneshot` allows multiple `ExecStart`s, `ExecStart` has to be a list. Even though multiple `ExecStarts` aren't allowed in other service types, the directive's type is still the same. Just like other lists, an empty assignment will clear them.

euank:
What @crawford's saying also matches my understanding.

lucab:
Ack, sorry for the noise.

PR #1024, opened 2017-03-23
*: replaced ignition and cloud config with container linux configs

dgonyeo:
Here's an incomplete pass at changing all the OS docs over to using container linux configs instead of ignition or cloud configs.

Everything alphabetically before `iscsi.md` should be good to go, with the exception of `cluster-architecture.md`, which needs a complete rewrite.

PR #1038, opened 2017-04-10
os: remove extra clcs

crawford:
This looks like it was leftover from testing how markdown handled
indented code blocks.


dgonyeo:
Looking at the diff in my PR these were added by mistake. Not sure how that happened.

https://github.com/coreos/docs/pull/1024/files#diff-4360a051ecdd7c0104f9c29a4dace7a7L265

PR #1043, opened 2017-04-17
os: add provisioning documentation

crawford:


os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+$ bin/ct < example.yml
+error: platform must be specified to use templating
+```
+
+This says that the config takes advantage of templating (e.g. `PUBLIC_IPV4`) but ct was invoked without being told what environment to target.
+
+CT can be invoked again and given Amazon EC2 as an example:
+
+```
+$ bin/ct --platform=ec2 < example.yml
```
ElijahCaine: If the platform *cannot* be set in the yaml config, explicitly state that. if the feature is slated to be added eventually, say that too.

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+This Ignition Config enables and configures etcd as specified in the above Container Linux Config. This can be more easily seen if the contents of the etcd drop-in are formatted nicely:
+
+```ini
+[Unit]
+Requires=coreos-metadata.service
+After=coreos-metadata.service
+
+[Service]
+EnvironmentFile=/run/metadata/coreos
+ExecStart=
```
ElijahCaine: Is this supposed to be here?

crawford: Yes. `ExecStart` is a list in systemd, so every assignment appends something to that list. A non-zero list is only valid for service types of `oneshot`, so we clear the list by assigning null.

os/provisioning.md
crawford: The platform cannot be explicitly stated in the config (that would defeat the purpose of making it platform independent). What gave you impression that this might be a feature that would be supported?

os/img/transpile-workflow.svg
```diff
@@ -0,0 +1 @@
...6,2.56,0,0,1-.82.38,4.08,4.08,0,0,1-1.11.14,4.51,4.51,0,0,1-1.59-.27,5.75,5.75,0,0,1-1.27-.64Z" fill="#54a3d9"/><path d="M103.19,287.83a1,1,0,0,1,.28-.74,1,1,0,0,1,1.38,0,1,1,0,0,1,.28.74,1,1,0,0,1-.28.73,1,1,0,0,1-1.38,0A1,1,0,0,1,103.19,287.83Zm0,4.3a1,1,0,0,1,.28-.74,1,1,0,0,1,1.38,0,1,1,0,0,1,.28.74,1,1,0,0,1-.28.73,1,1,0,0,1-1.38,0A1,1,0,0,1,103.19,292.13Z" fill="#54a3d9"/><path d="M80,303.06h5.16v.74H80Z" fill="#54a3d9"/><path d="M94.48,301.56h.82l.08,1h0a4.68,4.68,0,0,1,1-.82,2.27,2.27,0,0,1,1.21-.32,1.86,1.86,0,0,1,1.53.59,2.73,2.73,0,0,1,.49,1.76v3.62h-1V303.9a2,2,0,0,0-.31-1.23,1.23,1.23,0,0,0-1-.4,1.7,1.7,0,0,0-.49.07,1.86,1.86,0,0,0-.44.2,3.33,3.33,0,0,0-.44.33q-.23.2-.48.46v4.07h-1Z" fill="#54a3d9"/><path d="M101.54,305.88a1.56,1.56,0,0,1,.23-.86,1.86,1.86,0,0,1,.74-.63,5.22,5.22,0,0,1,1.31-.43,16.66,16.66,0,0,1,1.93-.26,2.13,2.13,0,0,0-.1-.56,1.23,1.23,0,0,0-.26-.46,1.29,1.29,0,0,0-.47-.32,1.88,1.88,0,0,0-.71-.12,2.83,2.83,0,0,0-1.1.22,6.85,6.85,0,0,0-.94.47l-.38-.67a5.45,5.45,0,0,1,.51-.29,5.66,5.66,0,0,1,.62-.27,5.26,5.26,0,0,1,.7-.2,3.69,3.69,0,0,1,.76-.08,2.27,2.27,0,0,1,1.78.65,2.53,2.53,0,0,1,.58,1.75v3.58h-.8l-.08-.79h0a5.63,5.63,0,0,1-1.09.66,2.87,2.87,0,0,1-1
```
joshix: etcd-member.service, perhaps? (Let's avoid adding new etcd2 mentions.)
I don't really grok the manual-vs-automatic distinction. It seems like Matchbox should be represented as "just another" provisioning target; maybe even at the top as the sort of "native target" vs the provider targets selected by their respective command line option

crawford: Ah, I should have called this out. We are going to use a different config and fix the capitalization across the board.

The manual vs automatic distinction is trying to show that manual "pipelines" are the most painful way to use CLCs. Tools like Matchbox (and eventually Terraform) will transparently transpile the config for the end user.

os/provisioning.md
ElijahCaine: I'd say something like that that (platform has to be passed as a flag because the config is OS agnostic).

This misconception came from the expectation that you'd be able to pass the platform in the yaml instead of being *required* to pass a CLI flag. I intuited that it *could* be set in the config file and after combing through the doc that expectation was broken.

Perhaps we could set a default parameter in the yaml and have the CLI [optionally] override that default? That might jive better with people's expectations. That said I accept that I may be in the minority here. This isn't the time nor place for feature requests, just throwing ideas out.

crawford: Addressed.

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+
+On top of this variable-expansion problem, it is very common for users to inadvertently write invalid configs. Instead of forcing users to waste time testing potentially invalid configs, the use of a transformation tool is instead encouraged. The default tool recommended for this task is the [Config Transpiler][ct] (ct for short). It is responsible for validating and transforming a Container Linux Config into the format that Container Linux can consume: the Ignition Config.
+
+[clc]: https://coreos.com/os/docs/latest/configuration.html
```
joshix: reference style links are ideally all collected just above EOF -- see coreos/docs/STYLE.md

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+[ct]: https://github.com/coreos/container-linux-config-transpiler/releases
+
+### Ignition Config
```
joshix: probably ## rather than ### I think

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+
+The Ignition Config is the configuration format consumed directly by Container Linux. Ignition, the utility inside of Container Linux which is responsible for provisioning the machine, actually fetches and executes the Ignition Config.
```
joshix: s/inside of/in/
s/which is responsible/responsible/

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+
+Ignition Configs are mostly static, distro-agnostic, and meant to be generated by a machine rather than a human. While they can be written directly by users, it is highly discouraged due to the easy that which errors can be introduced. Instead of writing Ignition Configs directly, users are encouraged to use provisioning tools like Matchbox, which transparently translate Container Linux Configs to Ignition Configs, or use the Config Transpiler directly.
```
joshix: link to Matchbox - https://coreos.com/matchbox/

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+
+![visual overview of the alternate ct workflows](img/transpile-workflow.svg)
+
+As can be seen above, ct will typically only be manually invoked when users are manually provisioning machines. If a provisioning tool like Matchbox is used, ct will transparently be incorporated into the deployment pipeline. In this scenario, the end user only needs to worry about preparing a Container Linux Config - Ignition and the Ignition are merely an implementation detail and knowledge of their existence isn't necessary.
```
joshix: s/Ignition and the Ignition are/Ignition and the Ignition Config are/
s/merely//
s/ and knowledge of their existence isn't necessary//

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Config Transpiler
+
+The Container Linux Config Transpiler is the tool that makes it possible for the end user to remain distanced from the details of configuring Container Linux. It's responsible for transforming a Container Linux Config written by an end user into an Ignition Config to be consumed by instances of Container Linux.
```
joshix: FIrst sentence would be better maybe like "The Container Linux Config Transpiler abstracts the details of configuring Container Linux."
s/end user/user/ (or "human", even) (throughout this line)

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+```container-linux-config:none
+etcd:
+  advertise_client_urls:       http://{PUBLIC_IPV4}:2379
+  initial_advertise_peer_urls: http://{PRIVATE_IPV4}:2380
+  listen_client_urls:          http://0.0.0.0:2379
+  listen_peer_urls:            http://{PRIVATE_IPV4}:2380
+  discovery:                   https://discovery.etcd.io/<token>
+```
+
+It was alluded to earlier that ct requires information about the target environment before it can transform configs which use templating. If this config is passed to ct without any other arguments, ct fails:
```
joshix: `ct` throughout line

os/provisioning.md
```diff
@@ -0,0 +1,127 @@
...
+ExecStart=
+ExecStart=/usr/lib/coreos/etcd-wrapper $ETCD_OPTS \
+  --listen-peer-urls="http://${COREOS_EC2_IPV4_LOCAL}:2380" \
+  --listen-client-urls="http://0.0.0.0:2379" \
+  --initial-advertise-peer-urls="http://$${COREOS_EC2_IPV4_LOCAL}:2380" \
+  --advertise-client-urls="http://${COREOS_EC2_IPV4_PUBLIC}:2379" \
+  --discovery="https://discovery.etcd.io/<token>"
+```
+
+The details of these changes are covered in depth in Ignition's [metadata documentation][metadata], but the gist is that `coreos-metadata` is used to fetch the IP addresses from the Amazon APIs and then systemd is leveraged to substitute the IP addresses into the invocation of etcd. The result is that even though Ignition only runs once, `coreos-metadata` fetches the IP addresses whenever etcd is run, allowing etcd to use IP addresses that have the potential to change.
```
joshix: `systemd`

os/provisioning.md
crawford: Ah, I've been throwing them at the end of their section.

PR #1055, opened 2017-04-24
docs: 1st draft editorial pass

zbwright:
Just a language pass.

### DIFF
--- a/os/provisioning.md
+++ b/os/provisioning.md
@@ -4,7 +4,9 @@ Container Linux automates machine provisioning with a specialized system for app
 
 ## Container Linux Config
 
-Container Linux admins define these configurations in a format called the [Container Linux Config][clc]. Container Linux Configs are structured as YAML and intended to be human-readable. The Container Linux Config is specific to Container Linux, as the name implies. It has features devoted to configuring Container Linux services such as [etcd][etcd], [rkt][rkt], Docker, [flannel][flannel], and [locksmith][locksmith]. The defining feature of the Container Linux Config is that it cannot be sent directly to a Container Linux provisioning target. Instead, it is first validated and transformed into a machine-readable and wire-efficient form. Before examining why and how that process is conducted, a few example Container Linux Configs will give a flavor of the format:
+Container Linux admins define these configurations in a format called the [Container Linux Config][clc], which is specific to Container Linux, structured as YAML, and intended to be human-readable. The Container Linux Config has features devoted to configuring Container Linux services such as [etcd][etcd], [rkt][rkt], Docker, [flannel][flannel], and [locksmith][locksmith]. The defining feature of the config is that is cannot be sent directly to a Container Linux provisioning target. Instead, it is first validated and transformed into a machine-readable and wire-efficient form.
+
+The following examples demonstrate the simplicity of the Container Linux Config format.
 
 This extremely simple Container Linux Config will fetch and run the current release of etcd:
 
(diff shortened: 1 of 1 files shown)


os/provisioning.md
```diff
@@ -4,22 +4,24 @@ Container Linux automates machine provisioning with a specialized system for app
...
+Container Linux admins define these configurations in a format called the [Container Linux Config][clc], which is specific to Container Linux, structured as YAML, and intended to be human-readable. The Container Linux Config has features devoted to configuring Container Linux services such as [etcd][etcd], [rkt][rkt], Docker, [flannel][flannel], and [locksmith][locksmith]. The defining feature of the Config is that is cannot be sent directly to a Container Linux provisioning target. Instead, it is first validated and transformed into a machine-readable and wire-efficient form.
```
crawford: "feature of the Config is that" should be lowercase.

os/provisioning.md
```diff
@@ -30,25 +32,27 @@ etcd:
...
+`PUBLIC_IPV4` and `PRIVATE_IPV4` are automatically populated from the environment (such as EC2, Azure, or GCE) in which Container Linux runs, if this metadata exists. For example: For EC2 (the default), the `public_ipv4` and `local_ipv4` metadata would be used. In Azure, *either* the virtual IP or public IP could be used for the `PUBLIC_IPV4` and the dynamic IP would be used for the `PRIVATE_IPV4`. On bare metal, this information cannot be reliably derived in a general manner, so these variables cannot be used.
```
crawford: EC2 isn't the default. That statement is referring to the bare-bones config we give as an example in the EC2 docs (default is probably not the right word).

crawford: The purpose of this paragraph is to walk the user through the edge cases which drove the decision to require the platform to be specified. This rewrite doesn't make that obvious and makes the Azure case very confusing.

zbwright: I see. so, questions: This doc should walk users through the standard implementation, which it does for most of the doc. 1) Are all three of these platforms edge cases? and, more importantly 2) when you say 'are populated' what you mean is 'Populate these variables according to the environment in which CL runs.' Correct? I wasn't paying enough attention to the code, apparently. I saw curly brackets, and didn't think variable. 3) Also, any reason not to use one standard for variables - either {} or <> ?

zbwright: @crawford to flesh out a little:
(maybe add a sub-heading: Public and Private IPV4 addresses.)
1. This is how it works.
2. There are some ambiguities and murky waters.
3. This is what you do in those cases.
4. Here's how to check your work.

os/provisioning.md
```diff
@@ -32,9 +32,9 @@ etcd:
...
+`PUBLIC_IPV4` and `PRIVATE_IPV4` are automatically populated from the environment in which Container Linux runs, if this metadata exists. Given the many different environments in which Container Linux can run, it's difficult if not impossible to accurately determine these variables in every instance. Be certain to check this value as a troubleshooting measure.
 
-Given the many different environments in which Container Linux can run, it's difficult if not impossible to accurately determine these variables in every instance.
+For example, the default metadata for an EC2 environment would be used: `public_ipv4` and `local_ipv4`. On Azure, *either* the virtual IP or public IP could be used for the `PUBLIC_IPV4` and the dynamic IP would be used for the `PRIVATE_IPV4`. On bare metal, this information cannot be reliably derived in a general manner, so these variables cannot be used.
```
crawford: On Azure, *either* the virtual IP or public IP could be used for the `PUBLIC_IPV4` (`ct` makes a best guess and uses the virtual IP, but this could change in the future) and the dynamic...

PR #1066, opened 2017-04-27
os/booting-on-azure: Update for ARM deployment model and Azure CLI 2.0

bgilbert:
The current Azure CLI doesn't support the classic deployment model. Drop obsolete exposition, as well as command-line arguments with reasonable defaults.

### DIFF
--- a/os/booting-on-azure.md
+++ b/os/booting-on-azure.md
@@ -16,19 +16,19 @@ The following command will create a single instance. For more details, check out
     <div class="tab-pane" id="alpha">
       <div class="channel-info">
         <p>The Alpha channel closely tracks master and is released frequently. The newest versions of system libraries and utilities will be available for testing. The current version is Container Linux {{site.alpha-channel}}.</p>
-        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data config.ign --image CoreOS:CoreOS:Alpha:{{site.alpha-channel}}</pre>
+        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data "$(cat config.ign)" --image CoreOS:CoreOS:Alpha:{{site.alpha-channel}}</pre>
       </div>
     </div>
     <div class="tab-pane" id="beta">
       <div class="channel-info">
         <p>The Beta channel consists of promoted Alpha releases. The current version is Container Linux {{site.beta-channel}}.</p>
-        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data config.ign --image CoreOS:CoreOS:Beta:{{site.beta-channel}}</pre>
+        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data "$(cat config.ign)" --image CoreOS:CoreOS:Beta:{{site.beta-channel}}</pre>
       </div>
     </div>
     <div class="tab-pane active" id="stable">
       <div class="channel-info">
         <p>The Stable channel should be used by production clusters. Versions of Container Linux are battle-tested within the Beta and Alpha channels before being promoted. The current version is Container Linux {{site.stable-channel}}.</p>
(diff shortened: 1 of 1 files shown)


os/booting-on-azure.md
```diff
@@ -16,19 +16,19 @@ The following command will create a single instance. For more details, check out
     <div class="tab-pane" id="alpha">
       <div class="channel-info">
         <p>The Alpha channel closely tracks master and is released frequently. The newest versions of system libraries and utilities will be available for testing. The current version is Container Linux {{site.alpha-channel}}.</p>
-        <pre>azure vm create --custom-data=config.ign --vm-size=Small --ssh=22 --ssh-cert=path/to/cert --no-ssh-password --vm-name=node-1 --location="&lt;location&gt;" my-cloud-service $(azure vm image list --json | jq --raw-output '.[].name | select(contains("__CoreOS-Alpha-{{site.alpha-channel}}"))') core</pre>
+        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data config.ign --image CoreOS:CoreOS:Alpha:{{site.alpha-channel}}</pre>
```
crawford: Does `az` use the image labels that allow you to omit the version number? Ideally, we'd use those and just specify that users should run the latest Alpha rather than a specific version which happens to be the latest.

bgilbert: Yes, it does.  Will change.

os/booting-on-azure.md
```diff
@@ -59,55 +59,34 @@ etcd:
...
-Let's create two more instances:
-
-```sh
-azure vm create --custom-data=config.ign --ssh=2022 --ssh-cert=path/to/cert --no-ssh-password --vm-name=node-2 --connect=my-cloud-service $(azure vm image list --json | jq --raw-output '.[].name | select(contains("__CoreOS-Stable-{{site.alpha-channel}}"))') core
-```
+Now that you have a resource group, create an instance of Container Linux Alpha {{site.alpha-channel}} inside it:
 
 ```sh
-azure vm create --custom-data=config.ign --ssh=3022 --ssh-cert=path/to/cert --no-ssh-password --vm-name=node-3 --connect=my-cloud-service $(azure vm image list --json | jq --raw-output '.[].name | select(contains("__CoreOS-Stable-{{site.alpha-channel}}"))') core
+az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data "$(cat config.ign)" --image CoreOS:CoreOS:Alpha:{{site.alpha-channel}}
```
crawford: Needs `--ssh-key-value` as well.

os/booting-on-azure.md
```diff
@@ -16,19 +16,19 @@ The following command will create a single instance. For more details, check out
     <div class="tab-pane" id="alpha">
       <div class="channel-info">
         <p>The Alpha channel closely tracks master and is released frequently. The newest versions of system libraries and utilities will be available for testing. The current version is Container Linux {{site.alpha-channel}}.</p>
-        <pre>azure vm create --custom-data=config.ign --vm-size=Small --ssh=22 --ssh-cert=path/to/cert --no-ssh-password --vm-name=node-1 --location="&lt;location&gt;" my-cloud-service $(azure vm image list --json | jq --raw-output '.[].name | select(contains("__CoreOS-Alpha-{{site.alpha-channel}}"))') core</pre>
+        <pre>az vm create --name node-1 --resource-group group-1 --admin-username core --custom-data "$(cat config.ign)" --image CoreOS:CoreOS:Alpha:{{site.alpha-channel}}</pre>
```
crawford: Use `CoreOS:CoreOS:Alpha:latest` as the URN instead.

os/booting-on-azure.md
bgilbert: The ssh key is automatically pulled from ~/.ssh/id_rsa.pub.  If that file is missing, `az` will fail with a message:

    An RSA key file or key value must be supplied to SSH Key Value. You can use --generate-ssh-keys to let CLI generate one for you

So I think it's safe to omit.

crawford: Ah, interesting. I was running it within a container, so I didn't get the shorthand. I agree, we should omit it.

PR #1076, opened 2017-05-16
os/booting-with{i,]pxe.md: mention supported schemas

dgonyeo:
This commit adds text mentioning the supported url schemas (http(s) and
tftp) to the pxe and ipxe booting guides.

Prompted by ignition getting tftp support: https://github.com/coreos/ignition/pull/353

Should probably not be merged until the next Ignition release.


bgilbert:
To avoid the double "URL", what about e.g. "`http`, `https`, and `tftp` schemes are supported"?

dgonyeo:
That seems more sensible. Updated.

PR #1079, opened 2017-05-19
Quay Enterprise: RBAC on tectonic documentation for 1.6

alekssaul:
Updated documentation for QE on Tectonic 1.6

Tectonic 1.5 manifests works if kubectl version 1.6 is used. During this `kubectl` seem to transform `apiVersion: rbac.authorization.k8s.io/v1alpha1` to `apiVersion: rbac.authorization.k8s.io/v1beta1` . 

Created 1.6 specific manifests for cases where user may not be using kubectl or an older version of kubectl.

### DIFF
--- a/quay-enterprise/tectonic/files/quay-servicetoken-role-k8s1-6.yaml
+++ b/quay-enterprise/tectonic/files/quay-servicetoken-role-k8s1-6.yaml
@@ -1,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:  
+  name: quay-enterprise-serviceaccount
+  namespace: quay-enterprise
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - put
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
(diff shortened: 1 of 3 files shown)


alekssaul:
@ericchiang / @josephschorr

alekssaul:
Maybe it makes sense to drop Kubernetes 1.4 Manifests?

quay-enterprise/tectonic/files/quay-servicetoken-role-k8s1-6.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - secrets
+  verbs:
+  - get
+  - put
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
```
ericchiang: On a "Role" this isn't going to work since namespaces aren't a namespaced resource. What's the actual requirement here? Being able to see the namespace it's working in?

You do get the ability to get your own namespace just by creating this role though (that's a bit hidden in the RBAC docs). Is that enough?

alekssaul: Hi Eric,
I've taken the RBAC roles for 1.5 and changes the `apiVersion` from rbac alpha to beta. From Quay's perspective, it needs to have access to modify a secret object that gets created in the quay-enterprise namespace which QE stores it's configuration at.

ericchiang: Okay. Lets just merge as is then. We can clean it up later if we care.

ericchiang:
lgtm

alekssaul:
@joshix : ok to merge?

PR #1087, opened 2017-05-24
Revert "Revert "os/booting-with{i,]pxe.md: mention supported schemas""

bgilbert:
We'll need this when Ignition &ge; 0.15.0 has re-merged to master.  **Do not merge** until then.

This reverts #1086.


bgilbert:
`oem` needs more context, which is [here](https://github.com/bgilbert/docs/blob/a291edfe6111112cd61159f7ae04d5d8d899afe5/os/booting-with-pxe.md#adding-a-custom-oem).  `data` doesn't seem like a good thing to recommend on the kernel command line.

PR #1094, opened 2017-06-06
booting-with-libvirt.md: Removed invalid argument

mikalv:
virt-install's parameter --network don't support any "type=bridge" argument, removing it makes the vm boot.


mikalv:
Forgot to mention OS and version;
Ubuntu 16.04
package: virtinst - 1:1.3.2-3ubuntu1.16.04.3
virt-install --version
1.3.2

PR #1100, opened 2017-06-15
sshd: update de-socket-activationizing instructions to be a little nicer

euank:
See individual commit comments.

cc @crawford and @kbrwn

### DIFF
--- a/os/customizing-sshd.md
+++ b/os/customizing-sshd.md
@@ -77,6 +77,7 @@ Since [Container Linux Configs][cl-configs] are only applied on first boot, exis
 
 The following sections walk through applying the same changes documented above on a running machine.
 
+*Note*: To avoid incidentally locking yourself out of the machine, it's a good idea to double-check you're able to directly login to the machine's console, if applicable.
 
 ### Changing the sshd port
 
--- a/os/customizing-sshd.md
+++ b/os/customizing-sshd.md
@@ -34,7 +34,7 @@ storage:
 
 ### Changing the sshd port
 
-Container Linux ships with socket-activated SSH by default. The configuration for this can be found at `/usr/lib/systemd/system/sshd.socket`. We're going to override some of the default settings for this in the Container Linux Config provided at boot:
+Container Linux ships with socket-activated SSH daemon by default. The configuration for this can be found at `/usr/lib/systemd/system/sshd.socket`. We're going to override some of the default settings for this in the Container Linux Config provided at boot:
 
 ```yaml container-linux-config
 systemd:
@@ -65,11 +65,11 @@ systemd:
     mask: true
 ```
 
-Note that in this configuration the port will be configured by updating the `/etc/ssh/sshd_config` file with the `Port` directive rather than via editing `sshd.socket`.
+Note that in this configuration the port will be configured by updating the `/etc/ssh/sshd_config` file with the `Port` directive rather than via `sshd.socket`.
 
 ### Further reading
 
-Read the [full Container Linux Config][cl-configs] guide for more details on working with Container Linux Configs, including setting uesr's ssh keys.
+Read the [full Container Linux Config][cl-configs] guide for more details on working with Container Linux Configs, including setting user's ssh keys.
 
 ## Customizing sshd after first boot
 
@@ -108,7 +108,7 @@ FreeBind=true
 
 You can specify both an IP and an alternate port in a single `ListenStream` line. IPv6 address bindings would be specified using the format `[2001:db8::7]:22`.
 
-*Note*: While a specific IP address is optional, you must always specify the port, even if it is the default SSH port. The FreeBind option is used to allow the socket to be bound on addresses that are not yet configured on an interface, to avoid issues caused by delays in IP configuration at boot. (This option is required only if you are specifying an address.)
+*Note*: While specifying an IP address is optional, you must always specify the port, even if it is the default SSH port. The `FreeBind` option is used to allow the socket to be bound on addresses that are not yet configured on an interface, to avoid issues caused by delays in IP configuration at boot. (This option is required only if you are specifying an address.)
 
 Multiple ListenStream lines can be specified, in which case `sshd` will listen on all the specified sockets:
 
@@ -123,7 +123,7 @@ FreeBind=true
 
 ### Activating changes
 
-After creating the the dropin file, the changes can be activated by doing a daemon-reload and restarting `sshd.socket`
+After creating the dropin file, the changes can be activated by doing a daemon-reload and restarting `sshd.socket`
 
 ```
 $ sudo systemctl daemon-reload
--- a/os/customizing-sshd.md
+++ b/os/customizing-sshd.md
@@ -50,7 +50,7 @@ systemd:
 
 `sshd` will now listen only on port 222 on all interfaces when the system is built.
 
-### Disabling socket-activation for sshd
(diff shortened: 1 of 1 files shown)


euank:
and cc @zbwright since this is os-docs

os/customizing-sshd.md
```diff
@@ -128,30 +163,25 @@ $ systemctl status sshd.socket
...
+Container Linux by CoreOS stable (1353.8.0)
+core@machine $
 ```
 
-### Further reading on systemd units
-
-For more information about configuring Container Linux hosts with `systemd`, see [Getting Started with systemd](getting-started-with-systemd.md).
-
-### Override socket-activated SSH
+### Disabling socket-activation for sshd
```
zbwright: socket activation

os/customizing-sshd.md
```diff
@@ -41,38 +41,73 @@ systemd:
...
+### Changing the sshd port
+
+First, create an editable copy of the `sshd.socket`.
 
 ```
 $ sudo cp /usr/lib/systemd/system/sshd.socket /etc/systemd/system/sshd.socket
 ```
 
-This gives you a copy of the sshd.socket unit that will override the one supplied by Container Linux -- when you make changes to it and `systemd` re-reads its configuration, those changes will be implemented.
+This gives you a `sshd.socket` unit that will override the one supplied by Container Linux -- when you make changes to it and `systemd` re-reads its configuration, those changes will be applied.
```
zbwright: ... by Container Linux. When changes are made to it, and `systemd` ...

zbwright:
and one more:

change 
You can specify both an IP and an alternate port in a single ListenStream line; additionally, IPv6 address bindings would be specified as, for example, [2001:db8::7]:22 Note two things: first, while a specific IP address is optional, you must always specify the port, even if it is the default SSH port. Second, the FreeBind option is used allow the socket to be bound on addresses that are not actually configured on an interface yet, to avoid issues caused by delays in IP configuration at boot. (This is not needed if you are not specifying an address.)

to 
You can specify both an IP and an alternate port in a single ListenStream line. IPv6 address bindings would be specified using the format [2001:db8::7]:22. 
Note: While a specific IP address is optional, you must always specify the port, even if it is the default SSH port. The FreeBind option is used to allow the socket to be bound on addresses that are not yet configured on an interface, to avoid issues caused by delays in IP configuration at boot. (This option is required only if you are specifying an address.)

os/customizing-sshd.md
```diff
@@ -41,38 +41,73 @@ systemd:
...
+To configure sshd on Container Linux without socket activation, a Container Linux Config file similar to the following may be used:
+
+```yaml container-linux-config
+systemd:
+  units:
+  - name: sshd.socket
+    mask: true
+  - name: sshd.service
+    enable: true
+    contents: |-
```
crawford: Does our existing service not have an install section? We should just fix that so that we don't need to overwrite the unit here.

os/customizing-sshd.md
```diff
@@ -41,38 +41,73 @@ systemd:
       enable: true
       contents: |
         [Socket]
-        ListenStream=2222
+        ListenStream=222
```
crawford: Why the port change?

euank: See the commit message https://github.com/coreos/docs/pull/1100/commits/3c184a0d0c1db8eb3e220db1fe8466b75135f794

os/customizing-sshd.md
euank: After looking properly at the unit we ship, I have no clue why we're overwriting it.

I'm pretty sure I started overwriting it just because our existing docs said to and never checked properly.

With fresh eyes, it looks like masking `sshd.socket` is sufficient. I'll update it to that.

euank:
Properly updated to be everything-is-a-dropin.

I manually verified the first section works by launching libvirt machines with each ignition snippet.

euank:
cc @crawford @zbwright, I think I addressed all comments.

os/customizing-sshd.md
```diff
@@ -9,8 +9,10 @@ As a practical example, when a client fails to connect by not completing the TCP
 ## Customizing sshd with a Container Linux Config
 
 In this example we will disable logins for the `root` user, only allow login for the `core` user and disable password based authentication. For more details on what sections can be added to `/etc/ssh/sshd_config` see the [OpenSSH manual][openssh-manual].
+If you're interested in additional security options, Mozilla provides a well-commented example of a [hardened configuration][mozilla-ssh-rec].
 
 [openssh-manual]: http://www.openssh.com/cgi-bin/man.cgi?query=sshd_config
+[mozilla-ssh-rec]: https://wiki.mozilla.org/Security/Guidelines/OpenSSH#Modern_.28OpenSSH_6.7.2B.29
```
zbwright: we typically put the [name]: link at the bottom of the page, after two blank lines.

os/customizing-sshd.md
```diff
@@ -128,88 +137,36 @@ $ systemctl status sshd.socket
...
-   Listen: [::]:2222 (Stream)
-           10.20.30.40:2223 (Stream)
+   Listen: [::]:222 (Stream)
+           10.20.30.40:223 (Stream)
  Accepted: 1; Connected: 0
 ...
 ```
 
-And if we attempt to connect to port 22 on our public IP, the connection is rejected, but port 2222 works:
+And if we attempt to connect to port 22 on our public IP, the connection is rejected, but port 222 works:
```
zbwright: question - should the first port mentioned be 22, and the 2nd be 222 ?

euank: Isn't that what it is?

euank:
@crawford look okay to you on the technical side now?

os/customizing-sshd.md
```diff
@@ -23,11 +23,11 @@ storage:
           # Use most defaults for sshd configuration.
           UsePrivilegeSeparation sandbox
           Subsystem sftp internal-sftp
+          UseDNS no
 
           PermitRootLogin no
           AllowUsers core
-          PasswordAuthentication no
-          ChallengeResponseAuthentication no
```
dm0-: I feel like we've been through this before.  My system's `sshd_config` says about `UsePAM`:

> If you just want the PAM account and session checks to run without
> PAM authentication, then enable this but set PasswordAuthentication
> and ChallengeResponseAuthentication to 'no'.

euank: We have been through this before. In practice it seems this change is necessary though regardless of what docs say.

euank: Ah, no, you're right. This change isn't strictly necessary. Any reason I should prefer the previous config over this though? If not, I'm inclined to keep it as a whitelist.

dm0-: I would prefer to follow what is documented, since I don't know if there are unintended consequences of the change.

euank: My preference comes from Mozilla's config (https://wiki.mozilla.org/Security/Guidelines/OpenSSH#Modern_.28OpenSSH_6.7.2B.29) which is at least a form of documentation.

Based on the comments in that, it looks like these options are older and thus more widely supported. Fortunately, we don't have to support openssh 5.3

dm0-: It looks okay.  I suppose anyone needing something else not in the whitelist (e.g. GSSAPI) probably won't be referencing this example anyway.

PR #1104, opened 2017-06-19
os/sdk-modifying-coreos: removed repo and cros_sdk instructions

sdemos:
The "option 2" section of the sdk instructions is apparently obsolete. cc @euank

### DIFF
--- a/os/sdk-modifying-coreos.md
+++ b/os/sdk-modifying-coreos.md
@@ -27,9 +27,9 @@ git config --global user.name "Your Name"
 
 **NOTE**: Do the git configuration as a normal user and not with sudo.
 
-### Option 1: Using Cork
+### Using Cork
 
-The `cork` utility, included in the CoreOS [mantle](https://github.com/coreos/mantle) project, can be used to create and work with an SDK chroot.
+The `cork` utility, included in the CoreOS [mantle](https://github.com/coreos/mantle) project, is used to create and work with an SDK chroot.
 
 In order to use this utility, you must additionally have the [golang](https://golang.org/) toolchain installed and configured correctly.
 
@@ -47,7 +47,7 @@ export PATH=$PATH:$HOME/bin
 You may want to add the `PATH` export to your shell profile (e.g. `.bashrc`).
 
 
-Next, the cork utility can be used to create an SDK chroot:
+Next, use the cork utility to create a project directory. This will hold all of your git repos and the SDK chroot. A few gigabytes of space will be necessary.
 
 ```sh
 mkdir coreos-sdk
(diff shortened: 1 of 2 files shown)


bgilbert:
The option 1 text has two instances of `can be` that can be strengthened now that cork is the only option.  There's also one remaining reference to `cros_sdk` further down the page.

Also also, there are `cros_sdk` references in `sdk-building-production-images` and `sdk-tips-and-tricks` that we should clean up at some point.

sdemos:
I can modify the `cros_sdk` references in `sdk-building-production-images` and `sdk-tips-and-tricks` in this pr as well, if it is entirely deprecated.

euank:
Yeah, let's get rid of it everywhere we can.

euank:
ping @sdemos

sdemos:
Sorry, forgot about this pr.

Question for the changes in `sdk-building-productions-images` - is `COREOS_SDK_VERSION` used anywhere else? The line with `cros_sdk` in there seems to imply that `cros_sdk` is the only thing it's needed by. 

Question for the changes in `sdk-tips-and-tricks` - is there another way to "recreate the chroot prior to a clean rebuild" other than `./chromite/bin/cros_sdk -r`?

@bgilbert / @euank for answers. That should entirely remove any references to that tool.

euank:
In that context, `COREOS_SDK_VERSION` is still valid/relevant. `cork create` will still bootstrap from an existing sdk version, which is still derived from that.

For the second one, I think `cork create --replace` does the same thing?

sdemos:
This pr should be all set now.

os/sdk-modifying-coreos.md
```diff
@@ -61,57 +61,6 @@ cork enter
...
-export PATH="$PATH:$HOME/bin"
-curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
-chmod a+x ~/bin/repo
-```
-
-You may want to add the `PATH` export to `.bashrc` or `/etc/profile.d/` so that you don’t need to set `$PATH` in every new shell.
-
-#### Bootstrap the SDK chroot
-
-Create a project directory. This will hold all of your git repos and the SDK chroot. A few gigabytes of space will be necessary.
```
bgilbert: This information is missing from the option 1 text; it'd be good to move it over.

os/sdk-modifying-coreos.md
```diff
@@ -61,57 +61,6 @@ cork enter
...
-
-#### Use cros\_sdk to setup the chroot
-
-Download and enter the SDK chroot which contains all of the compilers and tooling.
-
-```sh
-./chromite/bin/cros_sdk
-```
-
-**WARNING:** To delete the SDK chroot, use `./chromite/bin/cros_sdk --delete`. Otherwise, you will delete `/dev` entries that are `bind`-mounted into the chroot.
```
bgilbert: Does cork need a similar warning?

sdemos: I'm not sure, but cork does have `cork delete` which will delete the sdk chroot for you, so I'll include that information in the same style.

sdemos:
@bgilbert are the new changes satisfactory?

os/sdk-modifying-coreos.md
bgilbert: It turns out that `cork delete` just wraps `sudo rm -rf`, so let's drop the warning for now.

bgilbert:
We now have a [published cork binary](https://github.com/coreos/mantle/releases).  Could you update the instructions to cover downloading and GPG-verifying the binary, rather than building from source?

sdemos:
updated to download the cork binary and verify. /cc @bgilbert

os/sdk-modifying-coreos.md
```diff
@@ -31,17 +31,15 @@ git config --global user.name "Your Name"
...
 First, install the cork utility:
 
 ```sh
-git clone https://github.com/coreos/mantle
-cd mantle
-./build cork
-mkdir ~/bin
-mv ./bin/cork ~/bin
-export PATH=$PATH:$HOME/bin
+curl https://github.com/coreos/mantle/releases/download/v0.7.0/cork-0.7.0-amd64
```
bgilbert: Does not follow GitHub's redirect (`-L`), and if it did, would dump the binary to the terminal.

os/sdk-modifying-coreos.md
```diff
@@ -31,17 +31,15 @@ git config --global user.name "Your Name"
...
+curl https://github.com/coreos/mantle/releases/download/v0.7.0/cork-0.7.0-amd64.sig
+gpg --verify cork-0.7.0-amd64.sig cork-0.7.0-amd64
```
bgilbert: We can't assume the user will already have imported the signing key, and we should be explicit about what output to expect.

os/sdk-modifying-coreos.md
sdemos: hah turns out I definitely didn't test these commands.

os/sdk-modifying-coreos.md
```diff
@@ -61,56 +61,7 @@ cork enter
...
-#### Use cros\_sdk to setup the chroot
-
-Download and enter the SDK chroot which contains all of the compilers and tooling.
-
-```sh
-./chromite/bin/cros_sdk
-```
-
-**WARNING:** To delete the SDK chroot, use `./chromite/bin/cros_sdk --delete`. Otherwise, you will delete `/dev` entries that are `bind`-mounted into the chroot.
+**WARNING:** To delete the SDK chroot, use `cork delete`. Otherwise, you will delete `/dev` entries that are `bind`-mounted into the chroot.
```
bgilbert: Can be dropped per earlier discussion.

sdemos: should this happen before I merge it?

bgilbert: Yeah.

PR #1105, opened 2017-06-20
torcx: add metadata file docs

lucab:


lucab:
/cc @squeed @dm0- to have a look at this

Assigning to @joshix for further dispatching

dm0-:
As mentioned in Slack, it might be worth explicitly noting that `torcx.target` isn't started by default, so something must depend on it before its status is active.  All units installed by torcx images require it.

I'm not against enabling it by default if you want to allow for cases where users disable the vendor profile (and its pre-enabled `docker.socket`) and unpack torcx images that only provide binaries and no units.  That would be a case where `torcx.target` isn't started on boot, but users might still want to check on its status.

/cc @euank

lucab:
Yes, thanks for the feedback. I assumed it was a glitch in the ebuild and wrote this doc under that perspective, but I'll amend it to reflect current status quo.

I think either ways are fine, as long as we document what is the expected usage. I'm slightly more in favor of a target enabled by default, as it makes easier spotting generator failures.

lucab:
Second iteration: amended to mention that `torcx.target` is disabled by default and should be brought in as a dependency.

torcx/metadata-and-systemd-target.md
```diff
@@ -0,0 +1,76 @@
+# Metadata and systemd target
+
+In many cases, it is desirable to inspect the state of a system booted with Torcx and to verify the details of the configuration that has been applied.
+For these purposes, Torcx comes with additional facilities to integrate with systemd-based workflows: a custom target and a metadata file containing environment flags.
```
joshix: s/these purposes/this purpose/

torcx/metadata-and-systemd-target.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Metadata entries and environment flags
+
+In order to signal a successful run, Torcx writes a metadata file at most once per boot. The format of this file is suitable for consumption by systemd `EnvironmentFile=` [directive][systemd-exec] and can be used to introspect the booted configuration at runtime.
```
joshix: s/by systemd/by the systemd/

torcx/metadata-and-systemd-target.md
```diff
@@ -0,0 +1,76 @@
...
+  "kind": "profile-manifest-v0",
+  "value": {
+    "images": []
+  }
+}
+```
+
+## Torcx target unit
+
+System services may depend on successful execution of Torcx generator. As such, `torcx.target` is provided as a target unit which is only reached if the generator run without failures and the system got successfully sealed.
```
joshix: s/run/runs/
s/got/was/

torcx/metadata-and-systemd-target.md
```diff
@@ -0,0 +1,76 @@
...
+
+This target is not enabled by default, but can referenced as a dependency by other units who want to introspect system status:
```
joshix: s/can referenced/can be referenced/

torcx/metadata-and-systemd-target.md
lucab: Shouldn't these two verb forms match? I.e. both either present or past forms

lucab: Reworded the sentence following @squeed's suggestion to avoid the empasse and turning it into a simpler form.

lucab:
Third iteration: typos spotted by @joshix.

PR #1114, opened 2017-06-28
Confusing "Config Transpiler" documentation

naisanza:
The readme at https://github.com/coreos/container-linux-config-transpiler is a lot more clearer than the docs, which just led to confusion.


bgilbert:
Hi @naisanza, that link should really point to a specific Markdown doc.  That way, when `installing-to-disk.md` is rendered for our docs website, it will be converted to a link within the site.  I've made that change in https://github.com/coreos/docs/pull/1186.

Thanks for the PR!

PR #1115, opened 2017-06-29
torcx: add troubleshooting guide

lucab:
/cc @squeed

### DIFF
--- a/torcx/troubleshooting.md
+++ b/torcx/troubleshooting.md
@@ -1,0 +1,40 @@
+## Troubleshooting
+
+Torcx generator runs early in the boot, when other system facilities are not yet set up and available for use. In case of errors, troubleshooting and debugging can be performed following the suggestions described here.
+
+### Checking for failures
+
+In case of errors, Torcx stops before sealing the new system state. This means that in order to check for correct execution, it is sufficient to verify that the metadata file exists:
+
+```
+$ test -f /run/metadata/torcx || echo 'torcx failed'
+```
+
+On failures, the metadata seal file will not exist, and `torcx failed` will be printed. Verify failure at boot time using the `torcx.target` unit:
+
+```
+$ sudo systemctl start torcx.target ; sudo systemctl status torcx.target
+
+Assertion failed on job for torcx.target.
+
+* torcx.target - Verify torcx succeeded
+   Loaded: loaded (/usr/lib/systemd/system/torcx.target; disabled; vendor preset: disabled)
+   Active: inactive (dead) since [...]
+   Assert: start assertion failed at [...]
+           AssertPathExists=/run/metadata/torcx was not met
+```
+
+### Gathering logs
+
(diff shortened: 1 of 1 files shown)


lucab:
Assigning to @joshix for further dispatching.

torcx/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
+## Troubleshooting
+
+Torcx generator runs early in the boot, when other system facilities are not yet set up and available for use. In case of errors, troubleshooting and debugging can be performed following the suggestions described here.
+
+### Checking for failures
+
+In case of errors, Torcx stops before being able to seal the new system state. This means that in order to check for correct execution, it is sufficient to verify that the metadata file exists:
```
squeed: Should read "Torcx stops before sealing..."

lucab:
Second iteration: small language nits.

lucab:
Gentle bump for a review.

torcx/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+### Gathering logs
+
+The single most useful piece of information needed when troubleshooting failure is the log from `torcx-generator`. This binary does not run as a typical systemd service, thus log filtering must be done via its syslog identifier.
+With systemd-journald, this can be accomplished with the following command:
+
+```
+$ journalctl --boot 0 --identifier /usr/lib64/systemd/system-generators/torcx-generator
+```
+
+In the event that this doesn't yield any results, running as root may help. There are circumstances where the journal isn't owned by the systemd-journal group or the current user is not a part of that group.
```
zbwright: If this doesn't yield results, run as root. There may be instances in which the journal isn't owned by the systemd-journal group, or the current user is not part of that group.

torcx/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Validating the configuration
+
+One common cause for Torcx failures is a malformed configuration (e.g. wrongly assembled profile or a syntax error). Less likely, the active profile could reference addon images which are no more available on the system.
```
zbwright: One common cause for Torcx failure is a malformed configuration (such as a mis-assembled profile, or a syntax error). In other cases, the active profile might reference addon images which are no longer available on the system. 

(then, leave a blank line before 'Use `torcx profile check` to validate a profile configuration.'

torcx/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Checking for failures
+
+In case of errors, Torcx stops before sealing the new system state. This means that in order to check for correct execution, it is sufficient to verify that the metadata file exists:
+
+```
+$ test -f /run/metadata/torcx || echo 'torcx failed'
+```
+
+On failures, the metadata seal file won't exist thus printing the "torcx failed" output. This can be also verified at boot-time using the `torcx.target` unit:
```
zbwright: On failures, the metadata seal fill will not exist, and `torcx failed` will be printed. Verify failure at boot time using the `torcx.target` unit.

lucab:
@zbwright thanks, reworded according to your suggestions (I dropped one stale paragraph too).

PR #1117, opened 2017-07-03
Added option to download create-coreos-vdi script using curl.

baig:


euank:
Apologies for not getting to this sooner.

I personally don't see the need to enumerate multiple download methods that are all basically equivalent.

If a reader wishes to use one or the other for a specific reason, that's fine, and easy enough to do.
I think the documentation reads better though when it is reasonably specific and prescriptive.

Do you think there's a specific reason curl might be preferred over wget?

PR #1120, opened 2017-07-08
os/swap: update instructions

euank:
The previous ones didn't use systemd `.swap` units, which are the
appropriate tool for this.

Note: do not merge until I've tested this, this is untested.

cc @dgonyeo @zbwright as some possible reviewers

### DIFF
--- a/os/adding-swap.md
+++ b/os/adding-swap.md
@@ -2,63 +2,62 @@
 
 Swap is the process of moving pages of memory to a designated part of the hard disk, freeing up space when needed. Swap can be used to alleviate problems with low-memory environments.
 
-By default Container Linux does not include a partition for swap, however one can configure their system to have swap.
+By default Container Linux does not include a partition for swap, however one can configure their system to have swap, either by including a dedicated partition for it or creating a swapfile.
 
 ## Managing swap with systemd
 
-A systemd-managed service can be used to create, enable, and disable a swapfile service on your system. Write, enable, and start the systemd unit file found in the next section to add swap to a Container Linux node.
+systemd provides a specialized `.swap` unit file type which may be used to activate swap. The below example shows how to add a swapfile and activate it using systemd.
+
+### Creating a swapfile
+
+The following commands, run as root, will make a 1GiB file suitable for use as swap.
+
+```sh
+mkdir -p /var/vm
+fallocate -l 1024m /var/vm/swapfile1
+chmod 600 /var/vm/swapfile1
+mkswap /var/vm/swapfile1
+```
 
 ### Creating the systemd unit file
 
-The following systemd unit creates and enables a service to manage a 1GiB swapfile. It should be written to `/etc/systemd/system/swap.service`.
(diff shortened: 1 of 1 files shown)


os/adding-swap.md
```diff
@@ -2,63 +2,61 @@
...
 [Install]
 WantedBy=multi-user.target
 ```
 
-**Note** Ensure the block device containing your swapfile is configured to auto-mount on boot.
-
 ### Enable the unit and start using swap
 
-The following command enables and starts the new `swap` service.
+After creating the unit, it can be enabled. Optionally, the `swappiness` value can be modified if your system has sufficient memory for most tasks.
```
zbwright: delete 'Optionally' - Use `systemctl` to enable the unit once created. The `swappiness` value may be modified if the system has sufficient memory.

os/adding-swap.md
```diff
@@ -2,63 +2,61 @@
...
-The following command enables and starts the new `swap` service.
+Use `systemctl` to enable the unit once created. The `swappiness` value may be modified if desired.
 
-```
-$ systemctl enable --now /etc/systemd/system/swap.service
+```sh
+$ systemctl enable --now var-vm-swapfile1.swap
+# Optionally
+$ echo 'vm.swappiness=10' | sudo tee /etc/sysctl.d/80-swappiness.conf
+$ /usr/lib/systemd/systemd-sysctl
```
bgilbert: Should this be starting the systemd service instead?

os/adding-swap.md
```diff
@@ -2,63 +2,61 @@
...
+systemd provides a specialized `.swap` unit file type which may be used to activate swap. The below example shows how to add a swapfile and activate it using systemd.
 
-### Creating the systemd unit file
+### Creating a swapfile
 
-The following systemd unit creates and enables a service to manage a 1GiB swapfile. It should be written to `/etc/systemd/system/swap.service`.
+The following commands will make a 1GiB file suitable for use as swap.
 
+```sh
+mkdir -p /var/vm
```
bgilbert: All of these commands implicitly assume root.  Do we want to say that?

os/adding-swap.md
```diff
@@ -2,63 +2,61 @@
...
+fallocate -l 1024m /var/vm/swapfile1
```
bgilbert: Missing chmod?

zbwright:
tiny bump @euank - did you want to finish this out?

os/adding-swap.md
euank: Added `, as root, ` to the line above. For the others (using systemd, writing a unit) I think we frequently assume root in other docs and it's a fairly clear error / systemd login prompt / whatever if you omit it.

os/adding-swap.md
euank: Good catch

euank:
Made the suggested changes and ran through the manual steps once more, they worked on a test aws instance.

euank:
I'm leaving it up to someone from the docs team to merge

PR #1126, opened 2017-07-26
[WIP] os/quickstart: remove fleet, restructure

euank:
This is a first pass at a new structure for this document.

I haven't filled in most of the words, I mostly just want any ideas
around whether this structure/format makes sense, what headings / topics
should be covered, etc.

The thing I'm least sure about is whether the idea of "three tools" is
the right way to present this, and whether service discovery vs
provisioning matters as a member of that list.

cc @crawford @zbwright Per above, mostly looking for high level feedback, which is why it's basically just headers and some TODO bullet lists right now.


zbwright:
looks useful to me, and more clearly defined than the existing quickstart.md. go forth and write!

euank:
I unfortunately never got back to this.
If anyone would like to pick this up, or start fresh on rewriting the quickstart and close this, please be my guest.

The quickstart certainly still needs to be rewritten to not link to fleet (and probably to reference Kubernetes as a possible next step), but I don't think I'll be the one to do it.

Bubblemelon:
Can I be assigned to this?

euank:
Container Linux is deprecated, and this PR is clearly well past stale.

PR #1137, opened 2017-08-24
sdk-modifying-coreos: Mention cork setup

glevand:


os/sdk-modifying-coreos.md
```diff
@@ -128,6 +128,15 @@ For Debian, Ubuntu, and other Debian based systems installing the following pack
 
     sudo apt-get install binfmt-support qemu-user-static
 
+##### Configuring using Cork
```
bgilbert: I don't think we need both this and the next section, since they cover the same set of distros and the user will already have cork (because of https://github.com/coreos/docs/pull/1104).  Let's just update the instructions in the next section, and maybe say that it writes `/etc/binfmt.d/qemu-aarch64.conf` in case the user wants to remove it again.

glevand:
Rebased to latest.
Just updated existing section as requested.

PR #1138, opened 2017-08-25
os: Drop some stale SDK docs

bgilbert:
cc @ajeddeloh


ajeddeloh:
We will need to update the "Modifying CoreOS" docs to link to the mantle github project instead of this removed page. Other than that, LGTM

bgilbert:
Good catch.  Updated.

ajeddeloh:
LGTM

PR #1141, opened 2017-08-25
os: add docs for having root on raid

ajeddeloh:
Depends on:
https://github.com/coreos/mantle/pull/690
https://github.com/coreos/mantle/pull/689
https://github.com/coreos/bootengine/pull/132

cc: @bgilbert

### DIFF
--- a/os/root-filesystem-placement.md
+++ b/os/root-filesystem-placement.md
@@ -1,0 +1,64 @@
+# Configuring Root Filesystem Placement
+Container Linux supports composite disk devices such as RAID arrays. If the root filesystem is placed on a composite device, special care must be taken to ensure Container Linux can find and mount the filesystem early in the boot process. GPT partition entries have a [partition type GUID](https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs) that specifies what type of partition it is (e.g. Linux filesystem); Container Linux uses special type GUIDs to indicate that a partition is a component of a composite device containing the root filesystem.
+
+## Root on RAID
+RAID enables multiple disks to be combined into a single logical disk to increase reliability and performance. To create a software RAID array when provisioning a Container Linux system, use the `storage.raid` section of a [Container Linux Config](https://coreos.com/os/docs/latest/provisioning.html). RAID components containing the root filesystem must have the type GUID `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`. All other RAID arrays must not have that GUID; the Linux RAID partition GUID `a19d880f-05fc-4d3b-a006-743f0f84911e` is recommended instead. See the [Ignition documentation](https://coreos.com/ignition/docs/latest/examples.html#create-a-raid-enabled-data-volume) for more information on setting up RAID for data volumes.
+
+### Overview
+To place the root filesystem on a RAID array:
+
+ * Create the component partitions used in the RAID array with the type GUID `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`.
+ * Create a RAID array from the component partitions.
+ * Create a filesystem labeled `ROOT` on the RAID array.
+ * Remove the `ROOT` label from the original root filesystem.
+
+### Example Container Linux Config
+This Container Linux Config creates partitions on `/dev/vdb` and `/dev/vdc` that fill each disk, creates a RAID array named `root_array` from those partitions, and finally creates the root filesystem on the array. To prevent inadvertent booting from the [original root filesystem](https://coreos.com/os/docs/latest/sdk-disk-partitions.html#partition-table), `/dev/vda9` is reformatted with a blank ext4 filesystem labeled `unused`.
+
+**Warning: This will erase both `/dev/vdb` and `/dev/vdc`.**
+```yaml container-linux-config
+storage:
+  disks:
+    - device: /dev/vdb
+      wipe_table: true
+      partitions:
+       - label: root1
+         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+    - device: /dev/vdc
+      wipe_table: true
+      partitions:
+       - label: root2
+         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+  raid:
+    - name: "root_array"
+      level: "raid1"
+      devices:
+        - "/dev/vdb1"
+        - "/dev/vdc1"
+  filesystems:
+    - name: "ROOT"
+      mount:
+        device: "/dev/md/root_array"
+        format: "ext4"
+        create:
+          options:
+            - "-L"
+            - "ROOT"
+    - name: "unused"
+      mount:
+        device: "/dev/vda9"
+        format: "ext4"
+        create:
+          options:
+            - "-L"
+            - "unused"
+          force: true
+```
+
+### Limitations
+
+ * Other system partitions, such as `USR-A`, `USR-B`, `OEM`, and `EFI-SYSTEM`, cannot be placed on a software RAID array.
+ * RAID components containing the root filesystem must be partitions on a GPT-partitioned device, not whole-disk devices or partitions on an MBR-partitioned disk.
+ * `/etc/mdadm.conf` cannot be used to configure a RAID array containing the root filesystem.
+ * Since Ignition cannot modify the type GUID of existing partitions, the default `ROOT` partition cannot be reused as a component of a RAID array. A future version of Ignition will support resizing the `ROOT` partition and changing its type GUID, allowing it to be used as part of a RAID array.
+
--- a/os/sdk-tips-and-tricks.md
+++ b/os/sdk-tips-and-tricks.md
@@ -203,3 +203,4 @@ e96281a6-d1af-4bde-9a0a-97b76e56dc57
 
 - CoreOS Root: 5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6
 - CoreOS Reserved: c95dc21a-df0e-4340-8d7b-26cbfa9a03e0
+- CoreOS Raid Containing Root: be9067b9-ea49-4f15-b4f6-f36f8c9e1818


os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
+# Setting up Root on Raid
+
+Partitions used in RAID arrays need to have the gpt partition type uuid `be9067b9-ea49-4f15-b4f6-f36f8c9e1818` in order to boot. Because Ignition does not current support modifying existing partitions
+without wiping the partition table, the default ROOT partition cannot be used since it has different partition type uuid. In order to have the root partition in a raid array:
+ * Create the partitions used in the RAID array with the gpt partition type uuid `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`
+ * Create a RAID array with the created partitions 
+ * Create a filesystem with on the raid array with the ROOT label
+ * Remove the ROOT label from the old root filesystem by wiping the existing filesystem or overwriting it
+
+## Example CLC
```
bgilbert: `Container Linux Config`

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+         number: 2
+         size: 256MiB
+         start: 0
+         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+  raid:
+    - name: "my_raid_array"
+      level: "raid1"
+      devices:
+        - "/dev/vdb1"
+        - "/dev/vdb2"
```
bgilbert: Let's not recommend, even implicitly, that the user RAID together two partitions on the same disk.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+
+```yaml container-linux-config
+storage:
+  disks:
+    - device: /dev/vdb
+      wipe_table: true
+      partitions:
+       - label: root1
+         number: 1
+         size: 256MiB
```
bgilbert: Not a realistic size for a root filesystem.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+         start: 0
+         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+       - label: root2
+         number: 2
+         size: 256MiB
+         start: 0
+         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
+  raid:
+    - name: "my_raid_array"
```
bgilbert: Stylistic nit: there are going to be installations that use this name verbatim, so let's pick a better name.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+            - "-L"
+            - "unused"
+          force: true
+```
+
+## Under the Hood
+Since Container Linux has an immutable initramfs, the initramfs must have a predetermined way for finding the root filesystem. If the root partition is in a RAID array, the raid array must be started in order to find the root filesystem. To avoid assembling unnecessary arrays (which may be assembled with other options from mdadm.conf), the partitions that are part of the root raid array must have the partition type uuid `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`. A udev rule in the initramfs autoassembles all partitions with that type uuid.
+
+## Future
+Ignition will support resizing and relabling partitions which allow the root partition to be resized and assigned a new type uuid. This will allow it to be reused as part of a raid array.
```
bgilbert: There might be some value to this, but only in a particular context: if we've explained the situation well enough that an informed user is now thinking "that's odd", then we can fill in details about why we're doing things that way and how we're fixing the process.  But until we're at that level of clarity, talking about future changes will only confuse matters further.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
+# Setting up Root on Raid
+
+Partitions used in RAID arrays need to have the gpt partition type uuid `be9067b9-ea49-4f15-b4f6-f36f8c9e1818` in order to boot. Because Ignition does not current support modifying existing partitions
```
bgilbert: In order to boot?  Not true for data volumes.

While we shouldn't spend a lot of time explaining background concepts like "type UUID", we shouldn't casually use terms that may be obscure to the target audience.  One well-placed sentence of explanation can go a long way.

Let's not emphasize the limitations too early.  The user primarily wants to know how to complete the task.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+without wiping the partition table, the default ROOT partition cannot be used since it has different partition type uuid. In order to have the root partition in a raid array:
```
bgilbert: [Paragraphs should not be linewrapped.](https://github.com/coreos/docs/blob/master/STYLE.md#line-wrapping-considered-harmful)

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+ * Create a filesystem with on the raid array with the ROOT label
+ * Remove the ROOT label from the old root filesystem by wiping the existing filesystem or overwriting it
+
+## Example CLC
+
+```yaml container-linux-config
+storage:
+  disks:
+    - device: /dev/vdb
+      wipe_table: true
```
bgilbert: Maybe we should warn the user that this will wipe `vdb`.

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
...
+        format: "ext4"
+        create:
+          options:
+            - "-L"
+            - "unused"
+          force: true
+```
+
+## Under the Hood
+Since Container Linux has an immutable initramfs, the initramfs must have a predetermined way for finding the root filesystem. If the root partition is in a RAID array, the raid array must be started in order to find the root filesystem. To avoid assembling unnecessary arrays (which may be assembled with other options from mdadm.conf), the partitions that are part of the root raid array must have the partition type uuid `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`. A udev rule in the initramfs autoassembles all partitions with that type uuid.
```
bgilbert: This reads like an explanation of a design decision.  That's not appropriate for user docs.  If someone was told that they need to configure a RAID, they shouldn't need to know what an immutable initramfs or an unnecessary array or udev is.

The user needs to know:

- What to expect for root RAID and how to set it up (it needs a special GUID, and mdadm.conf is not respected)
- What to expect for data RAID and how to set it up (automatically assembled during boot, subject to settings in mdadm.conf, and *must not* use the special GUID)

os/root-on-raid.md
ajeddeloh: `root_array` perhaps?

os/root-on-raid.md
```diff
@@ -0,0 +1,58 @@
+# Setting up Root on Raid
```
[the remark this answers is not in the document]
ajeddeloh: Could the limitations go under their own subheading? Then the blurb about not being able to resuse the current root partition could live there too.

os/root-on-raid.md
bgilbert: Sure.

os/root-on-raid.md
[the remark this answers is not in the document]
bgilbert: I think a subheading makes sense.

Re USR, the user's intent is to set up a system resilient to disk failures, so I think we have to mention it here.  On other Linux distros, root-on-RAID would completely solve that problem.

os/root-on-raid.md
ajeddeloh: Should we suggest they use the generic linux raid partition GUID or just not the special one? The [Ignition docs](https://coreos.com/ignition/docs/latest/examples.html) don't specify one

bgilbert: 1. We should fix the Ignition docs.
2. In the long run (hopefully not too long) the Ignition user docs should really be migrated to CLC docs, coreos/bugs#2089.
3. I don't think step-by-step instructions for data RAID are in scope, but we should have a quick summary of the differences and a link to the docs that currently exist.

ajeddeloh: So we should sugguest they use the generic linux raid one (otherwise it will default to linux filesystem). I agree that we shouldn't give step by step instructions for data RAID, but if we are explicitly saying "Don't use the special guid" we should probably tell them a sane value to use instead (in addition to linking to relevent docs).

bgilbert: Yes, the generic one.  Let's go ahead and provide the alternative GUID and see how it reads.

os/root-filesystem-tricks.md
```diff
@@ -0,0 +1,71 @@
+# Root Filesystem Tricks
+This document covers setting up the root filesystem encapsulated in virtual devices like RAID arrays. Since the root filesystem must be mounted by the initramfs, special care must be taken to ensure the initramfs can find and mount it correctly. GPT partition entries have a partition type GUID that is used to specify what type of partition it is (e.g. linux filesystem). Container Linux uses special type GUIDs to indicate a partition contains the root filesystem when it is encapulated by another system, like RAID.
```
ajeddeloh: Not thrilled with the wording of "encapsulated in virtual devices like RAID arrays".
Does the part about using the GPT partition type GUIDs belong here or is it too in depth for this doc?

os/root-filesystem-tricks.md
```diff
@@ -0,0 +1,71 @@
...
+Partitions used in the RAID array containing the root filesystem need to have the type GUID `be9067b9-ea49-4f15-b4f6-f36f8c9e1818` to ensure the initramfs can find the root partition within. In order to have the root partition in a RAID array:
+
+ * Create the partitions used in the RAID array with the GPT partition type GUID `be9067b9-ea49-4f15-b4f6-f36f8c9e1818`
+ * Create a RAID array with the created partitions 
+ * Create a filesystem with on the RAID array with the ROOT label
+ * Remove the ROOT label from the old root filesystem by wiping the existing filesystem or overwriting it
+
+### Example Container Linux Config
+This Container Linux Config creates partitions on /dev/vdb and /dev/vdc consuming the whole disk, then creates a RAID array using the partitions and creates the root filesystem on the array.
+*Warning: This will wipe both /dev/vdb and /dev/vdc.*
```
ajeddeloh: Whoops, should be *** not **

os/root-filesystem-tricks.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Limitations
+
+ * The USR partitions cannot be on a RAID array.
+ * Since Ignition cannot create partitions without first wiping a partition table, the existing root partition cannot be reused in a RAID array.
+ * Since `mdadm.conf` is on the root filesystem, it cannot be read until the root filesystem is mounted and therefore cannot be used to configure a RAID array containing the root filesystem.
+ * Because the initramfs uses the GPT type GUID to find RAID components with the root filesystem, any RAID component with the root filesystem must be partitioned, even if that means having one partition consume the entire disk.
+
+### Future
+Ignition will support resizing and relabling partitions which allow the root partition to be resized and assigned a new type GUID. This will allow it to be reused as part of a RAID array.
```
ajeddeloh: Should this move up be part of the Ignition limitation explanation or just be removed completely?

os/root-filesystem-placement.md
```diff
@@ -0,0 +1,64 @@
...
+        create:
+          options:
+            - "-L"
+            - "unused"
+          force: true
+```
+
+### Limitations
+
+ * Other system partitions, such as USR-A, USR-B, OEM, and EFI-SYSTEM, cannot be placed on a software RAID array.
```
bgilbert: Backquote partition labels?

os/root-filesystem-placement.md
```diff
@@ -0,0 +1,64 @@
...
+ * RAID components containing the root filesystem must be partitions on a GPT-partitioned device, not whole-disk devices or partitions on an MBR-partitioned disk.
+ * `/etc/mdadm.conf` cannot be used to configure a RAID array containing the root filesystem.
+ * Since Ignition cannot modify the type GUID of existing partitions, the default ROOT partition cannot be reused as a component of a RAID array. A future version of Ignition will support resizing the ROOT partition and changing its type GUID, allowing it to be used as part of a RAID array.
```
bgilbert: Likewise.

os/root-filesystem-placement.md
```diff
@@ -0,0 +1,64 @@
+# Configuring Root Filesystem Placement
+Container Linux supports composite disk devices such as RAID arrays. If the root filesystem is placed on a composite device, special care must be taken to ensure Container Linux can find and mount the filesystem early in the boot process. GPT partition entries have a [partition type GUID](https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs) that specifies what type of partition it is (e.g. Linux filesystem). Container Linux uses special type GUIDs to indicate that a partition is a component of a composite device containing the root filesystem.
```
bgilbert: The transition to the third sentence is abrupt.  Maybe join the third and fourth sentences with a semicolon, `, and`, etc., to make the flow clearer?

PR #1142, opened 2017-09-01
rm -r golang

jzelinskie:
This is not the recommended way to write Go projects and is pointless to
keep around.


euank:
@bgilbert as far as I could tell the docs weren't exposed publicly on the website anywhere. I don't think we have a replacement.

We do have the "template project"(https://github.com/coreos/template-project) for some boilerplate, but nothing Go specific.

We could presumably have a "golang" branch of the template project using glide, a makefile, etc as most of our projects do.

jzelinskie:
I think all the projects differ too much. Some things will want to use bazel if they need codegen, dep, glide, etc...

bgilbert:
They weren't exposed publicly, but AFAIK we now no longer have any CoreOS Go style guide.  Ideally the non-controversial parts of the README would remain available somewhere.

PR #1146, opened 2017-09-08
kubernetes: add doc on adding self-signed registry certificate to k8s…

kbrwn:
… worker nodes

### DIFF
--- a/kubernetes/add-registry-certificate-worker-nodes.md
+++ b/kubernetes/add-registry-certificate-worker-nodes.md
@@ -1,0 +1,91 @@
+# Adding Registry Certificate to Kubernetes Worker Nodes 
+
+Many enterprise's use self-signed certificates to protect container registries. On k8s this can present a problem as docker requires that self-signed certificates be nested under the `/etc/docker/certs.d` on every host that will pull from a registry with a self-signed certificate. 
+
+This restriction can be solved with a DaemonSet that copies the `ca.crt` file to the needed directory on each host. 
+
+A secret will be mounted into the DaemonSet as a file and then copied. This secret must include the base64 encoded contents of the root CA (pem format) used to sign the container registry certs.
+
+```bash
+$ cat rootCA.pem | base64 -w 0
+LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURWekNDQWorZ0F3SUJBZ0lKQUxRd3FGRWVpakdyTUEwR0NTcUdTSWIzRFFFQkN3VUFNRUl4Q3pBSkJnTlYKQkFZVEFsaFlNUlV3[...]
+```
+
+
+#### registry-secret.yaml 
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: registry-ca
+  namespace: kube-system
+type: Opaque
+data:
+  registry-ca: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURWekNDQWorZ0F3SUJBZ0lKQUxRd3FGRWVpakdyTUEwR0NTcUdTSWIzRFFFQkN3VUFNRUl4Q3pBSkJnTlYKQkFZVEFsaFlNUlV3[...]
+```
+
+Use kubectl to create the `registry-ca` secret: 
+
+```
+kubectl create -f registry-secret.yaml 
+```
+
+The following DaemonSet mounts the CA as the file `/home/core/registry-ca` and then copies this file to the `/etc/docker/certs.d/reg.example.com/ca.crt`. 
+
+Replace `reg.example.com` with the hostname of your container registry. 
+
+NOTE: This is a privileged container as the hostPath volume type requires for write privileges. 
+
+#### registry-ca-ds.yaml 
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: registry-ca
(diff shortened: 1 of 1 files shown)


kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
+# Adding Registry Certificate to Kubernetes Worker Nodes 
+
+Many enterprise's use self-signed certificates to protect container registries. On k8s this can present a problem as docker requires that self-signed certificates be nested under the `/etc/docker/certs.d` on every host that will pull from a registry with a self-signed certificate. 
+
+This restriction can be solved with a DaemonSet that copies the `ca.crt` file to the needed directory on each host. 
+
+A secret will be mounted into the DaemonSet as a file and then copied. This secret must include the base64 encoded contents of the root CA (pem format) used to sign the container registry certs.
+
+```bash
+$ cat rootCA.pem | base64 -w 0
```
euank: redundant cat:

```
$ base64 -w 0 rootCA.pem
```

kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
...
+
+Use kubectl to create the `registry-ca` DaemonSet: 
+
+```
+kubectl create -f registry-ca-ds.yaml
+```
+
+Checking for success can be accomplished by deploying a Pod or DaemonSet that pulls from the container registry. 
+
+As DaemonSets must have RestartPolicy equal to Always the registry-ca DaemonSet should be removed after succes:
```
euank: Shouldn't we just leave the DS so that all new nodes that join the cluster also get the certificate?

kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
...
+      labels:
+        name: registry-ca
+    spec:
+      containers:
+      - name: registry-ca
+        image: fedora
+        securityContext:
+          privileged: true
+        command: [ 'bash' ]
+        args: [ '-c', 'cp /home/core/registry-ca /etc/docker/certs.d/reg.example.com/ca.crt && sleep 500' ]
```
euank: `sleep 500` isn't great.

If we're planning to sleep "forever" then use `exec sleep inf`. If you want to be more portable than fedora (e.g. be able to run on busybox) then you can do `exec tail -f /dev/null`.

kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
...
+  template:
+    metadata:
+      labels:
+        name: registry-ca
+    spec:
+      containers:
+      - name: registry-ca
+        image: fedora
+        securityContext:
+          privileged: true
```
euank: Is this really needed?

kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
...
+    k8s-app: registry-ca
+spec:
+  template:
+    metadata:
+      labels:
+        name: registry-ca
+    spec:
+      containers:
+      - name: registry-ca
+        image: fedora
```
euank: Why not something smaller, like alpine or busybox?

kubernetes/add-registry-certificate-worker-nodes.md
```diff
@@ -0,0 +1,91 @@
...
+        securityContext:
+          privileged: true
+        command: [ 'bash' ]
+        args: [ '-c', 'cp /home/core/registry-ca /etc/docker/certs.d/reg.example.com/ca.crt && sleep 500' ]
+        volumeMounts:
+        - name: etc-docker
+          mountPath: /etc/docker/certs.d/reg.example.com
+          readOnly: false
```
euank: that's the default, no need to specify it.

kubernetes/add-registry-certificate-worker-nodes.md
kbrwn: Hmmm, I didn't really consider this. I guess if I change the `sleep 500` to `sleep infinity` the pods won't get restarted.

kubernetes/add-registry-certificate-worker-nodes.md
kbrwn: Pretty much only because I don't like alpine.

kubernetes/add-registry-certificate-worker-nodes.md
kbrwn: I think so. 

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

```
the directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged container or modify the file permissions on the host to be able to write to a hostPath volume
```

kubernetes/add-registry-certificate-worker-nodes.md
joshix: That laugh is not necessarily an endorsement. ;)

kubernetes/add-registry-certificate-worker-nodes.md
euank: That's true in the general case due to user namespaces.

However, I don't think tectonic supports user-namespaces so I don't think there's any case where it's needed in a tectonic cluster

kbrwn:
closing in favor of: https://github.com/coreos/tectonic-docs/pull/27

PR #1151, opened 2017-09-15
Add DNS configuration necessary for proper static config

fahlmant:
Without some DNS server, this static config won't allow a host to access much of anything. This config can be verified  [here](https://www.freedesktop.org/software/systemd/man/systemd.network.html#[Network]%20Section%20Options)

### DIFF
--- a/os/network-config-with-networkd.md
+++ b/os/network-config-with-networkd.md
@@ -17,6 +17,7 @@ Name=enp2s0
 [Network]
 Address=192.168.0.15/24
 Gateway=192.168.0.1
+DNS=192.168.10
 ```
 
 Place the file in `/etc/systemd/network/`. To apply the configuration, run:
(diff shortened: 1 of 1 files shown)


os/network-config-with-networkd.md
```diff
@@ -17,6 +17,7 @@ Name=enp2s0
 [Network]
 Address=192.168.0.15/24
 Gateway=192.168.0.1
+DNS=192.168.10
```
euank: Should that be `192.168.0.10`?

There's also the `10-static.network` a few lines further down which would need the same treatment.

We could also use `DNS=1.2.3.4` to match the other example below's sample value and make it even more obvious the user should fill in the correct one.

fahlmant: Oop, nice catch. 

Didn't see that example below, I'll change this to `DNS=1.2.3.4` for consistency's sake.

fahlmant:
@euank Updated

PR #1157, opened 2017-10-03
docs: add kubernetes-networking.md

zbwright:

PR #1158, opened 2017-10-03
Remove k8s

zbwright:
these two docs lived here, rather than in coreos-kubernetes. removing them is part of the final cleanup.


joshix:
@zbwright both this and #1157 add the file `kubernetes/kubernetes-networking.md`. Does this 1158 supersede 1157?

zbwright:
yes. This is a result of my poor git skills. This supersedes #1157. I will close that PR.

PR #1163, opened 2017-10-18
docs: add dev guide landing page

zbwright:


os/developer-guides.md
```diff
@@ -0,0 +1,18 @@
+# Developer Guides
+
+Most users will never have to build Container Linux from source or modify it in any way. If you have a need to modify Container Linux, we provide an SDK that allows you to build your own developer images. We also provide OEM functionality for cloud providers and other companies that must customize Container Linux to run within their environment.
+
+* [Modifying Container Linux][mod-cl]
```
euank: Maybe we should toss in a link to mantle, or the kola subsection of it (https://github.com/coreos/mantle/#kola).

The 'modifying CL' page links to cork, so the other big mantle piece that could be useful is kola, which would be `[Testing Container Linux](https://github.com/coreos/mantle/#kola)`.

os/developer-guides.md
```diff
@@ -0,0 +1,20 @@
...
+* [Kola integration testing framework][mantle-utils]
+
+
+[sdk-tips]: sdk-tips-and-tricks.md
+[disk-layout]: sdk-disk-partitions.md
+[production-images]: sdk-building-production-images.md
+[mod-cl]: sdk-modifying-coreos.md
+[dev-images]: sdk-building-development-images.md
+[kernel-modules]: kernel-modules.md
+[mantle-utils]: https://github.com/coreos/mantle/#kola/
```
euank: The trailing slash here breaks the anchor link

euank:
cc @bgilbert, does this look like a good start / bag of links?

os/developer-guides.md
zbwright: oof. I was told trailing slash required for repo links - due to the doc sync process. Maybe we should link to README.md#kola instead?

euank: I'm surprised that's necessary here, especially given that this url wont' be rewritten.

Linking to README.md would be okay though.

bgilbert:
LGTM.

PR #1165, opened 2017-10-23
docs: add comment conventions

zbwright:


joshix:
I think this is both too prescriptive, and yet not precise enough. It doesn't specify whether a comment should be above or below the line of code to which it refers; "inline" is imprecise in the context; the various comment delimiters for different languages and their escape character cousins are not (and probably cannot be) addressed.

I think we actually want to point contributors toward _not_ commenting in code blocks, but rather to explain simple code clearly in the English paragraphs preceding and surrounding it (whenever possible – which subtlety is why I think we can maybe live without this new rule, but if we must add it, we should conventionalize it toward the minimalism of the existing rules.)

zbwright:
yep, there's no way to address all possible languages. this was added specifically to state "inline or before" (which I can change to above, if you think it's more clear). Came out of a conversation with Rithu, to match engineering's standards.

I find inline comments useful in calling out differences between two similar snippets, specific parameter usage, required formatting techniques, and a few other details. They aren't used often, but when they are, should be standardized.

PR #1171, opened 2017-11-01
os: Developer documentation updates

bgilbert:
cc @euank for 1d1eb63.

### DIFF
--- a/os/notes-for-distributors.md
+++ b/os/notes-for-distributors.md
@@ -4,7 +4,11 @@
 
 Images of Container Linux alpha releases are hosted at [`https://alpha.release.core-os.net/amd64-usr/`][alpha-bucket]. There are directories for releases by version as well as `current` with a copy of the latest version. Similarly, beta releases can be found at [`https://beta.release.core-os.net/amd64-usr/`][beta-bucket] and stable releases at [`https://stable.release.core-os.net/amd64-usr/`][stable-bucket].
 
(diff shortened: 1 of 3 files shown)


os/notes-for-distributors.md
```diff
@@ -2,7 +2,7 @@
 
 ## Importing images
 
-Images of Container Linux alpha releases are hosted at `https://alpha.release.core-os.net/amd64-usr/`. There are directories for releases by version as well as `current` with a copy of the latest version. Similarly, beta releases can be found at `https://beta.release.core-os.net/amd64-usr/`.
+Images of Container Linux alpha releases are hosted at [`https://alpha.release.core-os.net/amd64-usr/`][alpha-bucket]. There are directories for releases by version as well as `current` with a copy of the latest version. Similarly, beta releases can be found at [`https://beta.release.core-os.net/amd64-usr/`][beta-bucket] and stable releases at [`https://stable.release.core-os.net/amd64-usr/`][stable-bucket].
 
 If you are importing images for use inside of your environment it is recommended that you import from the `current` directory. For example to grab the alpha OpenStack version of Container Linux you can import `https://alpha.release.core-os.net/amd64-usr/current/coreos_production_openstack_image.img.bz2`. There is a `version.txt` file in this directory which you can use to label the image with a version number.
```
euank: These recommendations should be updated to not be racy.

The right way to download a thing is to download version.txt and use the version from it to resolve non-current urls.

Otherwise, given current is a moving pointer and cloudflare caches, it's possible to get an image which is not the same version as version.txt said.

bgilbert: Fixed.

PR #1184, opened 2017-11-30
quay-enterprise: add env var needed for adding custom certs to qe on k8s

kbrwn:
customer submitted update


quay-enterprise/insert-custom-cert.md
```diff
@@ -87,4 +87,25 @@ Add an entry for the cert and paste the full base64 encoded string under the ent
...
+
+```
+      containers:
+      - name: quay-enterprise-app
+        image: quay.io/coreos/quay:v2.6.1
+        ports:
+        - containerPort: 80
+       env: 
+       - name: REQUESTS_CA_BUNDLE 
+	 value: /conf/stack/custom-cert.crt
```
josephschorr: Indentation is off

zbwright: I think back it all the way to the left, please. we don't typically indent snippets like this - just start fully outdented, and go from there. like so:
```
containers:
- name: quay-enterprise-app
  image: quay.io/coreos/quay:v2.6.1
  ports:
  - containerPort: 80
 env: 
 - name: REQUESTS_CA_BUNDLE 
value: /conf/stack/custom-cert.crt
```

PR #1187, opened 2017-12-12
os/*: update to CLC 2.1 syntax

bgilbert:


os/booting-on-ec2.md
```diff
@@ -152,8 +152,7 @@ storage:
     - mount:
         device: /dev/xvdb
         format: ext4
-        create:
-          force: false
+        wipe_filesystem: true
```
ajeddeloh: This changes meaning; is the original wrong?

bgilbert: See the commit message.

PR #1190, opened 2017-12-13
os/sdk-disk-partitions: correct partition 2 info

ajeddeloh:
Partition 2 is not unused; it is used by grub when booting from bios. Update the docs to reflect that.

### DIFF
--- a/os/sdk-disk-partitions.md
+++ b/os/sdk-disk-partitions.md
@@ -7,7 +7,7 @@ Container Linux is designed to be reliably updated via a [continuous stream of u
 | Number | Label      | Description                                                       | Partition Type        |
 |:------:|------------|-------------------------------------------------------------------|-----------------------|
 | 1      | EFI-SYSTEM | Contains the bootloader                                           | VFAT                  |
-| 2      | BIOS-BOOT  | This partition is reserved for future use                         | (none)                |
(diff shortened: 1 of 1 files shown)


os/sdk-disk-partitions.md
```diff
@@ -7,7 +7,7 @@ Container Linux is designed to be reliably updated via a [continuous stream of u
 | Number | Label      | Description                                                       | Partition Type        |
 |:------:|------------|-------------------------------------------------------------------|-----------------------|
 | 1      | EFI-SYSTEM | Contains the bootloader                                           | VFAT                  |
-| 2      | BIOS-BOOT  | This partition is reserved for future use                         | (none)                |
+| 2      | BIOS-BOOT  | Contains the second stages of GRUB for use when booting from BIOS | grub core.img         |
```
bgilbert: `stages`, plural?

ajeddeloh: It's the language used here: https://en.wikipedia.org/wiki/BIOS_boot_partition

Grub's documention is... sparse; I haven't found anything to support that language or counter it. Either way I don't think it matters much.

ajeddeloh: Found out a little bit more. `core.img` includes multiple subimages like `diskboot.img` and `kernel.img`. I suspect that's what the plurality on the wikipedia page is refering to.

see [grub's documentation](https://www.gnu.org/software/grub/manual/grub/html_node/Images.html) for more information.

PR #1200, opened 2018-01-24
[quay] add schema config for Quay Enterprise

IvanCherepov:
adds docs for schema

### DIFF
--- a/quay-enterprise/schema.md
+++ b/quay-enterprise/schema.md
@@ -1,0 +1,279 @@
+ # Schema for Quay Enterprise #
+
+ _Note: all fields are optional unless otherwise marked_
+
+* **authentication_type** [string] required: The authentication engine to use for credential authentication.
+  * **enum**: Database, LDAP, JWT, Keystone, OIDC.
+  * **Example**: `Database`
+* **buildlogs_redis** [object] required: Connection information for Redis for build logs caching.
+  * **host** [string] required: The hostname at which Redis is accessible.
+    * **Example**: `my.redis.cluster`
+  * **password** [string]: The password to connect to the Redis instance.
+      * **Example**: `mypassword`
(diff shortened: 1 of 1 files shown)


zbwright:
I don't think this will work the way the Quay team wants it to work. The ask is that we publish an html document listing all Quay Enterprise schema. @crawford I'm pretty certain we cannot load individual style.css files into the sync process, and I don't know what it would do with a raw html file (schema.html). Do you have any suggestions for how this should be handled?

crawford:
`sync` doesn't actually care about these files at all. It's Jekyll that processes them. Generally speaking though, I don't think we want rendered assets in this source repo.

zbwright:
yeah - me neither. @IvanCherepov let's talk this out in slack.

zbwright:
Ivan is going to see if he can't output to markdown instead.

zbwright:
Use this (loosely) as an example: https://coreos.com/os/docs/latest/configuration.html
* Do the params need to be all caps?
* Make "Schema for Quay Enterprise" header1 (# Schema for Quay Enterprise)
* Make each param text, rather than bullet (AUTHENTICATION_TYPE [string] required)
* Bump nested lists out a level

(I think this will make it easier to read. The Container Linux example gets too complicated to follow. I think if each param were simply text, rather than bulleted text, it would add line breaks and formatting that makes it easier to follow.)

zbwright:
let me know when this merges, and we'll get @athai to add it to the site. I think just drop in in under APIs under Reference on the landing page. Sound good, @josephschorr ?

josephschorr:
@zbwright It should probably be under the setup section

zbwright:
cool. thanks.

josephschorr:
@IvanCherepov Please regenerate given recent schema changes

PR #1203, opened 2018-02-05
os/sdk-modifying-coreos: update system requirements

bgilbert:
Add gpg. Drop python2 now that we only support cork.


bgilbert:
`sudo` is not universal.  Will add.

PR #1205, opened 2018-02-09
os/booting-on-ec2: update for eu-west-3 and cn-northwest-1

bgilbert:


euank:
Have you gone through the sync/pages song-and-dance to render html and verify it looks right?

bgilbert:
Yep.

PR #1231, opened 2018-05-11
os/booting-on-ecs: config to allow IAM Task Roles

kwyjibos:
Added the base config/rules for IAM Task Roles to match latest AWS documentation for other distributions.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html#enable_task_iam_roles


danspain-plutoflume:
Do I need to add any more info on this?

PR #1236, opened 2018-05-22
build_image without dev

Bubblemelon:
Removed the `dev` arg in Render the CoreOS Container Linux image section.

`./build_image dev` becomes `./build_image`

As suggested by @dgonyeo, dev image is unstable and not a recommended method.

### DIFF
--- a/os/sdk-modifying-coreos.md
+++ b/os/sdk-modifying-coreos.md
@@ -151,7 +151,7 @@ Build all of the target binary packages:
 
 #### Render the CoreOS Container Linux image
 
-Build an image based on the binary packages built above, including development tools:
+Build the production image based on the binary packages built above:
 
 ```sh
 ./build_image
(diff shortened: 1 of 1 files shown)


os/sdk-modifying-coreos.md
```diff
@@ -154,7 +154,7 @@ Build all of the target binary packages:
 Build an image based on the binary packages built above, including development tools:
```
bgilbert: Since we're now building the prod image, `including development tools` should be dropped.

Bubblemelon: Hi @bgilbert , thanks for catching that!

os/sdk-modifying-coreos.md
```diff
@@ -151,10 +151,10 @@ Build all of the target binary packages:
 
 #### Render the CoreOS Container Linux image
 
-Build an image based on the binary packages built above, including development tools:
+Build the production image based on the binary packages built above:
```
bgilbert: nit: `a production image` rather than `the production image`.

Bubblemelon: Got it ! :smile:

Issue #1235, opened 2018-05-21
Minor Improvement to Using Cork Section


Bubblemelon:
```
.
.
.
gpg --receive-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7
gpg --verify cork.sig cork
```
GPG versions [earlier than 2.1](https://forums.gentoo.org/viewtopic-p-8081732.html?sid=b1ace696054afe65dadb54ca1491fb06) will generate the following error:
`gpg: Invalid option "--receive-keys"` 

**I would like to add a snippet saying:**

Use following as alternative:
```
gpg2 --receive-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7
gpg2 --verify cork.sig cork
```
Using gpg2 will generate the same output as indicated by the immediate example as specified in the  docs:
```
gpg: Signature made Thu 19 Apr 2018 03:33:40 PM PDT
gpg:                using RSA key 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7
gpg: Good signature from "CoreOS Application Signing Key <security@coreos.com>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 18AD 5014 C99E F7E3 BA5F  6CE9 50BD D3E0 FC8A 365E
     Subkey fingerprint: 9CEB 8FE6 B4F1 E9E7 52F6  1C82 CDDE 268E BB72 9EC7

```

bgilbert:
`--keyserver keys.gnupg.net --recv-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7` will work with both `gpg` and `gpg2`.  (`gpg` appears not to have a default keyserver, thus the first option.)  Want to make a PR?

Bubblemelon:
@bgilbert you're right! I suggested using `gpg2` in case we wanted to keep to the `--receive-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7` command format. 

I could easily change the current:
`gpg --receive-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7`
to 
`gpg --keyserver keys.gnupg.net --recv-keys 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7`

Which will work for older gpg versions, and I'll include the alternative using gpg2 in the doc as well.

[...]

PR #1238, opened 2018-05-22
GPG Command Amendment

Bubblemelon:
Verify mantle section is changed in the following ways:
1. gpg command works for older versions of gpg
2. expected output from `gpg --verify` is updated


Bubblemelon:
This PR references issue #1235

PR #1242, opened 2018-05-24
Fix inconsistencies/typos in examples and text

polettix:
Mostly wrong IP addresses, also removing one stray element `node4` maybe from a too generous copy-and-paste.


lucab:
LGTM, thanks a lot for the cleanup!

PR #1247, opened 2018-06-25
Fix flannel unit-file url

Jaskaranbir:
The current documentation points to incorrect (old?) URL, some serious inconvenience for lazy people like me.


lucab:
Indeed. Thanks for the PR!

PR #1259, opened 2018-09-20
torcx: move under os

lucab:
This moves all torcx-related docs to `os/`, together with the rest
of Container Linux pages.


lucab:
This is a simple file moving pre-agreed with @bgilbert, so I'll just self-merge tomorrow unless he has any blocking feedback.

PR #1270, opened 2018-12-25
Remove leading zero on base-10 number.

### LINKED ISSUE #1269: Migration guide converts string "0644" to raw 0644; should not use octal syntax.
After transpilation it leads to `"mode" : 420` in the ignition config where previously was `permissions: "0644"` in the cloud-config.

asiplas:
Per https://yaml.org/type/int.html a yaml variable may “…have leading “0” to signal an octal base…”

original from_files section is 644—using 0644 in the migrated YAML ends up leaving 420 after it's transpiled.

Closes #1269

### DIFF
--- a/os/migrating-to-clcs.md
+++ b/os/migrating-to-clcs.md
@@ -243,7 +243,7 @@ storage:
   files:
     - filesystem: "root"
       path:       "/etc/hostname"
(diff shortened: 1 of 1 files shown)


asiplas:
Nevermind was confused.

PR #1273, opened 2019-05-15
os: document disabling SMT

bgilbert:


bgilbert:
Updated to note that the CLC will re-enable SMT on the second boot on Intel CPUs that aren't vulnerable.

PR #1280, opened 2019-09-19
Added Linode as a provider

Guaris:
This change adds Linode as a provider. Linode has supported CoreOS for a long time now: https://www.linode.com/distributions


bgilbert:
Thanks for the PR.  Container Linux doesn't ship with any support for Linode, even as a community-supported platform.  There is no official Linode image, no OEM ID or OEM support package, no support in Ignition or Afterburn, and no official documentation.  In addition, Linode apparently doesn't provide any mechanism for passing an Ignition config to a machine, making it impossible to configure a Container Linux node in the usual ways.  As a result, we can't recommend attempting to run Container Linux on Linode.

